PR #2, opened 2021-08-24
[ATOM-16291] AZSLc: SRG Layout Offset Incorrect When Struct Ends in f…

galibzon:
[ATOM-16291] AZSLc: SRG Layout Offset Incorrect When Struct Ends in float3x3

This is version 1.7.25.

The fix, when aligning according to dx12 rules, was to do nothing when Alignment was requested upon "struct"
ending.

Adjusted expected test results.

Validated all the changes with Tim Jones shader playground.

### DIFF
--- a/src/AzslcIntermediateRepresentation.cpp
+++ b/src/AzslcIntermediateRepresentation.cpp
@@ -586,7 +586,19 @@ namespace AZ::ShaderCompiler
             const auto layoutPacking = middleEndconfigration.m_packConstantBuffers;
             uint32_t startAt = Packing::AlignOffset(layoutPacking, 0, Packing::Alignment::asStructStart, 0, 0);
             uint32_t strideSize = CalculateSizeOfRootConstantsCB(rootConstantStructUid, middleEndconfigration.m_isRowMajor, layoutPacking);
-            uint32_t endOffset = Packing::AlignOffset(layoutPacking, strideSize, Packing::Alignment::asStructEnd, 0, 0);
+            uint32_t endOffset = 0;
+            switch (layoutPacking)
+            {
+                case Packing::Layout::DirectXPacking:          
+                case Packing::Layout::RelaxedDirectXPacking:
+                    // asStructEnd does nothing in DirectX, but in case of padding the root constant We force
+                    // asStructStart in order to get 16byte alignment.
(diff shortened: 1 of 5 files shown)


src/AzslcIntermediateRepresentation.cpp
```diff
@@ -586,7 +586,19 @@ namespace AZ::ShaderCompiler
             const auto layoutPacking = middleEndconfigration.m_packConstantBuffers;
             uint32_t startAt = Packing::AlignOffset(layoutPacking, 0, Packing::Alignment::asStructStart, 0, 0);
             uint32_t strideSize = CalculateSizeOfRootConstantsCB(rootConstantStructUid, middleEndconfigration.m_isRowMajor, layoutPacking);
-            uint32_t endOffset = Packing::AlignOffset(layoutPacking, strideSize, Packing::Alignment::asStructEnd, 0, 0);
+            uint32_t endOffset = 0;
+            switch (layoutPacking)
+            {
+                case Packing::Layout::DirectXPacking:
```
jeremyong-az, line 592: Minor nit: use `[[fallthrough]];`

moudgils (approved):

src/AzslcIntermediateRepresentation.cpp
galibzon: I don't think [[fallthrough]] is helpful in this case. There's a break after the first two cases.

jeremyong-az: It's not so much about this particular code snippet and much more about being able to actually enable this warning as an error across the codebase to flag cases where fallthrough occurs *inadvertently*.

galibzon: [[fallthrough]] is only useful in this case:
case A:
    some code; (without break or return)
case B:
    more code;
     break;

The example above can be improved with:
case A:
    some code; (without break or return)
    [[fallthrough]]
case B:
    more code;
     break;


My code is not that case. It is
case A:
case B:
      some code;
   break;

There's no need to use  [[fallthrough]], as no compiler complains because of two consecutive "case" statements.

jeremyong-az: Gotcha I misunderstood your comment, thanks for clarifying

jeremyong-az (approved):

PR #3, opened 2021-08-24
[ATOM-14595] ConstantsData::SetConstant<Matrix3x3> Stomps Data

galibzon:
This is version 1.7.26.

The new default behavior of AZSLc is:
    For all structs/classes/SRGs:
    All word sized primitive variables will be pre-padded
    by a "float2" variable if the preceding variable is a float3x3 or
    is a struct that ends in float3x3.

To skip the default behavior the command line argumment
"--skip-mat33-padding" is now available.

Added test suites to validate all cases and the produced
offsets have been validated with Tim Jones shader playground.


santorac (approved):

src/AzslcIntermediateRepresentation.cpp
```diff
@@ -628,6 +632,201 @@ namespace AZ::ShaderCompiler
         m_rootConstantStructUID = rootConstantStructUid;
     }
 
+    void IntermediateRepresentation::PadAllFloat3x3WhenFollowedByFourOrLessBytes(const MiddleEndConfiguration& middleEndconfigration)
+    {
+        //! Helper lambda to check if a symbol is a 3x3 matrix.
+        auto isMatrix3x3 = [&](const IdentifierUID& symbolUid)
```
jeremyong-az, line 638: Is the `&` capture needed here?

src/AzslcIntermediateRepresentation.cpp
```diff
@@ -628,6 +632,201 @@ namespace AZ::ShaderCompiler
...
+                continue;
+            }
+
+            const VarInfo* varInfo = kindInfo->GetSubAs<VarInfo>();
+            if (!varInfo)
+            {
+                continue;
+            }
+
+            if (!isWordSizedPrimitive(middleEndconfigration, varInfo))
```
jeremyong-az: Nit: `scalar` is the term I see more often. Primitive to me is like, a point, line or triangle element I think?

src/AzslcIntermediateRepresentation.cpp
```diff
@@ -628,6 +632,201 @@ namespace AZ::ShaderCompiler
...
+        {
+            // We can deduce the name of parent struct, class or SRG from the name of the field that should come AFTER
+            // the dummy float2.
+            auto parentName = GetParentName(insertBeforeThisUid.GetName());
+
+            // Define the name of the new dummy float2:
+            // Remark: it is possible that We end up adding more than one dummy padding variable
+            // to the same struct. This is why We have a static counter that will be helpful to avoid
+            // name collision.
+            static int s_numDummyFloats = 0;
```
jeremyong-az, line 787: Just for my understanding, why a static as opposed to just a locally declared counter?

src/AzslcIntermediateRepresentation.h
```diff
@@ -222,6 +222,62 @@ namespace AZ::ShaderCompiler
             return {};
         }
 
+        ///////////////////////////////////////////////////////////////////////
+        //! [GFX TODO] - ATOM-16287. Validate other matrix dimensions.
```
jeremyong-az, line 226: I remember reading somewhere that we should avoid references or links to internal JIRA tickets

src/AzslcIntermediateRepresentation.cpp
galibzon: Good point. Will change that.

src/AzslcIntermediateRepresentation.h
moudgils: We still want to track all the TODOs to a ticket. So we should really be using GHI (Github issue) numbers instead of Jira. Eventually we want to move out of Jira completely.

src/AzslcIntermediateRepresentation.cpp
galibzon: It uses "this", and fortunately "&" captures only what it needs by reference. It's the lazy & efficient approach, unless there are good reasons not to.

src/AzslcIntermediateRepresentation.cpp
galibzon: It would complicate the code by keeping a counter per scope, etc. Not worth the effort.

src/AzslcIntermediateRepresentation.h
galibzon: It is forbidden to use full URLs. Just the ticket number is fine.

src/AzslcIntermediateRepresentation.cpp
```diff
@@ -628,6 +632,201 @@ namespace AZ::ShaderCompiler
...
+
+            if (uid.GetName().find("(") != string::npos)
+            {
+                // It's a function argument. Skip.
+                continue;
+            }
+
+            // We only care if the previous VarInfo is one of:
+            // 1- A struct or class whose last member is a Matrix3x3.
+            // XOR
```
moudgils, line 732: did you mean to say OR?

src/AzslcIntermediateRepresentation.h
```diff
@@ -222,6 +222,62 @@ namespace AZ::ShaderCompiler
...
+        ///////////////////////////////////////////////////////////////////////
+        //! This function is useful to workaround a bug in SpirvCross when
```
moudgils: bug in **dxc** (not spirv-cross)

moudgils (approved):

src/AzslcIntermediateRepresentation.h
galibzon: Will change that.

src/AzslcIntermediateRepresentation.cpp
galibzon: In this case OR works too.

PR #4, opened 2021-08-25
[ATOM-2430] Better Error Message For Unknown SRG Field

galibzon:
With this fix, AZSLc fails when undefined SRG variables or functions
are detected during code emission.

BEFORE:
The shader pipeline would rely on DXC to detect the error,
which would be a non-informative error like:
"""
shadersource:561:23: error: use of undeclared identifier 'MaterialSrg'"
"""

NOW:
AZSLc itself detects the error and the message looks like:
"""
StandardPBR_ForwardPass.azslin(1031,24) : Emitter error #267:
Undefined ShaderResourceGroup symbol 'MaterialSrg::m_specularF0Factor'
"""

Added a couple of test cases where the error is expected to occur.


src/AzslcEmitter.cpp
```diff
@@ -1211,6 +1211,34 @@ namespace AZ::ShaderCompiler
         }
     }
 
+    void CodeEmitter::ValidateIsNotUndefinedSrgVariableOrThrow(antlr4::Token* token, TokenToAst::AstNode* nodeFromToken) const
```
santorac: That's pretty verbose, hehe. I think "ValidateSrgMemberIsDefined" would be sufficient :)

src/AzslcEmitter.cpp
```diff
@@ -1211,6 +1211,34 @@ namespace AZ::ShaderCompiler
...
+                throw AzslcEmitterException(EMITTER_UNEXPECTED_EXPRESSION, token, errorMessage);
+            }
+            // We only care to call out the error if the token has the name of an SRG.
+            QualifiedName qualifiedSymbolName(FormatString("/%s", token->getText().c_str()));
+            if (const IdAndKind* idAndKind = m_ir->GetIdAndKindInfo(qualifiedSymbolName))
+            {
+                const auto& [uid, kindInfo] = *idAndKind;
+                if (kindInfo.IsKindOneOf(Kind::ShaderResourceGroup))
+                {
+                    const string errorMessage = FormatString("\nUndefined ShaderResourceGroup symbol '%s'", qualifiedIdCtx->getText().c_str());
```
santorac: Why the extra "\n" at the beginning?

src/AzslcEmitter.h
```diff
@@ -189,6 +189,13 @@ namespace AZ::ShaderCompiler
         bool AlreadyEmittedFunctionDeclaration(const IdentifierUID& uid) const;
         bool AlreadyEmittedFunctionDefinition(const IdentifierUID& uid) const;
 
+        //! Helper function used during GetTextInStream().
+        //! The function assumes @nodeFromToken comes from @token.
+        //! It checks whether the token that is about to be written to the output
+        //! is not an undefined SRG field/variable.
+        //! Throws an exception if it is an undefined ShaderResourceGroup field/variable.
+        void ValidateIsNotUndefinedSrgVariableOrThrow(antlr4::Token* token, TokenToAst::AstNode* nodeFromToken) const;
```
santorac: Maybe use the tern "Member" instead of "Variable" since it could be a function right?

galibzon: I was thinking Symbol, but Member works. Will change to Member.

src/AzslcEmitter.cpp
galibzon: That's a useless left over. Will remove. Thanks for the catch.

src/AzslcEmitter.cpp
galibzon: Not sure why my brain chose NotUndefined... Horrible. Thanks for the tip.

galibzon: Actually, I recall why. 

The idea is that not only SRG symbols will go through that function. So technically it checks that it is Not and undefined Srg Symbol. Nonetheless it can be confusing.

moudgils (approved):
Other than the points raised by Chris this PR looks good to me

santorac (approved):

PR #5, opened 2021-08-31
[ATOM-16287] Validate Alignment Across All floatRxC

galibzon:
[ATOM-16287] Validate Alignment Across All floatRxC
Between vulkan & dx12

This is v1.7.28

This is the exhaustive solution to the SRG Layout data alignment problem
when there are
float or float2 variables
preceded by matrices or structs that end in matrices of type:
float2x2, float3x2, float4x2
In such case the "float" or "float2" must be prepadded by a float3
variable.

The other cases are when
"float" type variables
are preceded by matrices or structs that end in matrices of type:
float2x3, float3x3, float4x3
In such case the "float" must be prepadded by a float2 variables.


src/AzslcIntermediateRepresentation.cpp
```diff
@@ -794,36 +863,28 @@ namespace AZ::ShaderCompiler
...
             // ordered list of the struct/class/SRG.
-            KindInfo* parentKindInfo = GetKindInfo({ parentName });
-            assert(parentKindInfo);
-            const bool isStructOrClass = parentKindInfo->IsKindOneOf(Kind::Class, Kind::Struct);
             if (isStructOrClass)
             {
                 parentKindInfo->GetSubRefAs<ClassInfo>().InsertBefore(newVarUid, Kind::Variable, insertBeforeThisUid);
             }
-            else if (parentKindInfo->IsKindOneOf(Kind::ShaderResourceGroup))
+            else // if (parentKindInfo->IsKindOneOf(Kind::ShaderResourceGroup))
```
santorac: Delete commented out code.

santorac (changes requested):

src/AzslcIntermediateRepresentation.cpp
galibzon: It's supposed to be a comment :-)
Replaced with the comment:
// Due to similar check done above, it is a ShaderResourceGroup

santorac (approved):

galibzon:
Will close this pull request per our latest decision to mark it as an error for the user, proposing solutions and providing a command line argument to bypass error checking for this issue.

PR #7, opened 2021-09-20
[ATOM-16433] AZSLc: Revert Matrix3x3 padding.

galibzon:
[ATOM-16433] AZSLc: Revert Matrix3x3 padding.
Trigger Error When identifying alignment issues.
Provide option to bypass error.

This is v1.7.29

This is the exhaustive solution to validate the struct/class/SRG Layout data alignment problem
when:
1- There are float or float2 variables
preceded by matrices or structs that end in matrices of type:
float2x2, float3x2, float4x2
In such case the "float" or "float2" should be prepadded by a float3
variable and the solution is describbed in the error produced
by AZSLc.

2- The other cases are when "float" type variables
are preceded by matrices or structs that end in matrices of type:
float2x3, float3x3, float4x3 In such case the "float" must be prepadded by a float2 variables.
This solution is describbed in the error produced
by AZSLc.

Alternatively the user can skip this aligment error validation with
'--no-aligment-validation' command line option.

Error message example:
-------------------------------------------------------------
tests\Emission\AsError\matRC_padding.azsl(,)
: IR error #131: Detected potential alignment issues related with DXC
flag '-fvk-use-dx-layout':
- A 'float3' variable should be added before the variable 'm_f3' in
  'struct /A/SD2/m_f3' at Line number 21 of
  'D:\o3de-azslc\tests\Emission\AsError\matRC_padding.azsl'
- A 'float3' variable should be added before the variable 'm_f5' in
  'struct /A/SD2/m_f5' at Line number 23 of
    'D:\o3de-azslc\tests\Emission\AsError\matRC_padding.azsl'
-------------------------------------------------------------

[ATOM-16433] AZSLc: Revert Matrix3x3 padding.
Trigger Error When identifying alignment issues.
Provide option to bypass error.


src/AzslcIntermediateRepresentation.cpp
```diff
@@ -756,75 +785,124 @@ namespace AZ::ShaderCompiler
...
+            if (!lineInfo)
+            {
+                // When the LineDirectiveInfo* is null, it means We have detected a variable that was added
+                // by AZSLc itself. e.g. Root Constant padding, etc.
+                // In such case, this is not an issue We want to interfere with.
+                return {};
+            }
+            const auto originallineNumber = GetLineNumberInOriginalSourceFile(*lineInfo, lineOfDeclaration);
+
+            string solution = FormatString("- A 'float%d' variable should be added before the variable '%s' in '%s %s' at Line number %zu of '%s'\n",
```
jeremyong-az, line 882: Might be worth mentioning here that such extra padding would not be needed if we successfully upgrade DXC in the future

src/AzslcIntermediateRepresentation.cpp
```diff
@@ -756,75 +785,124 @@ namespace AZ::ShaderCompiler
...
+            {
+                typeName = "struct";
+            }
+            else
+            {
+                typeName = "ShaderResourceGroup";
+            }
+
+            // Let's get the line number where @insertBeforeThisUid was found in the flat AZSL file.
+            const auto * tmpThis = this; // To disambiguate which version of GetSymbolSubAs<> to call.
```
jeremyong-az, line 869: Sorry, do you mind explaining what you encountered that necessitated this?

src/AzslcIntermediateRepresentation.cpp
galibzon: I'm going to leave a TODO in the code to revisit once DXC is upgraded.

src/AzslcIntermediateRepresentation.cpp
galibzon: There's a const and non-const version of GetSymbolSubAs<>. VC++ was failing to compile because it couldn't figure out which one to use. Even though I was assigning the result to a "const auto *". I can see why, but it could have been equally easy for VC++ to use the const version.

Anyways, by assigning 'this' to a 'const auto *' it had no choice but to call the const version.

src/AzslcIntermediateRepresentation.cpp
jeremyong-az: Well, what I mean is, the message itself should be _very clear_ that any actual mitigation done in response to this message should be considered temporary and rolled back soon.

src/AzslcIntermediateRepresentation.cpp
jeremyong-az: I suspect the issue is that `this` is captured implicitly via the lambda capture context as `&` and the parent context has `this` referring to a non-const instance pointer (note that overload resolution is always argument-based and does not factor in return type, hence ADL = argument dependent lookup). Why not make this an actual helper member function instead of an inline-lambda? Stylistic preference I suppose?

src/AzslcIntermediateRepresentation.cpp
galibzon: Created github issue for this PR, and left this note:
        // GFX-TODO: https://github.com/o3de/o3de-azslc/issues/9
        // Upgrade DXC and if the alignment issues are fixed when using
        // -fvk-use-dx-layout then this function can be deleted.

src/AzslcIntermediateRepresentation.cpp
galibzon: You are right, ADL in action.
In the end not worth it to promote the lambda to a function because, as is, all the code is contained into a single block of code, and it will be easy to delete in the near future

src/AzslcIntermediateRepresentation.cpp
jeremyong-az: To be clear, currently the format string reads "A float%d variable should be added..."

It's the language I disagree with here. It really _shouldn't_ need to be added, but it currently is as a workaround. The error message implies that adding the padding is somehow "more correct" when in fact, it's actually incorrect but just needed temporarily. The information regarding the DXC upgrade should be surfaced to the user IMO because nobody would know to look in this code to find the comment communicating as such.

galibzon: All the "detected" errors are preceded by this banner:
            "Detected potential alignment issues related with DXC flag '-fvk-use-dx-layout'.\n"
            "Alternatively you can use the option '--no-alignment-validation' to void this error and compile as is.:\n");

santorac (approved):

jeremyong-az (approved):

PR #10, opened 2021-09-29
The HLSL file should preserve the original line numbers

galibzon:
This version 1.7.30

Added a new class NewLineCounterStream that can wrap a ostream instance.
All strings that are streamed into NewLineCounterStream are scanned for
the '\n' (new line) character. Each '\n' character found is counted.

With the help of NewLineCounterStream, which is instantiated as
CodeEmitter::m_out, We can appoximate the output line number of the
HLSL symbols to be as close as possible to the input line number from
the AZSL file for the corresponding symbols.

The line numbers are simply matched by adding new lines to the
output file only if the current line number of the symbol about to be
output is less than the original line number.

The idea is that later if there are errors in the HLSL file, when
compiled with other tools like DXC, the error line number will be very
close (+/-) 2 lines to the original line number.

### DIFF
--- a/src/AzslcEmitter.cpp
+++ b/src/AzslcEmitter.cpp
@@ -488,7 +488,7 @@ namespace AZ::ShaderCompiler
 
     void CodeEmitter::EmitStruct(const ClassInfo& classInfo, string_view structName, const Options& options)
     {
-        EmitNewLinesAsNecessary(classInfo.GetOriginalLineNumber());
+        EmitEmptyLinesToLineNumber(classInfo.GetOriginalLineNumber());
 
         const bool hasName = (structName.length() > 0);
         const auto tabs = "    ";
@@ -669,7 +669,7 @@ namespace AZ::ShaderCompiler
 
     void CodeEmitter::EmitEnum(const IdentifierUID& uid, const ClassInfo& classInfo, const Options& options)
     {
-        EmitNewLinesAsNecessary(classInfo.GetOriginalLineNumber());
+        EmitEmptyLinesToLineNumber(classInfo.GetOriginalLineNumber());
 
         const auto& enumInfo = get<EnumerationInfo>(classInfo.m_subInfo);
 
@@ -720,7 +720,7 @@ namespace AZ::ShaderCompiler
 
         AstFuncSig* node = funcSub.m_defNode ? funcSub.m_defNode : funcSub.m_declNode;
 
-        EmitNewLinesAsNecessary(funcSub.GetOriginalLineNumber(emitAsDefinition));
+        EmitEmptyLinesToLineNumber(funcSub.GetOriginalLineNumber(emitAsDefinition));
 
         EmitAllAttachedAttributes(uid);
 
@@ -856,7 +856,7 @@ namespace AZ::ShaderCompiler
 
     void CodeEmitter::EmitVariableDeclaration(const VarInfo& varInfo, const IdentifierUID& uid, const Options& options, VarDeclHasFlag declOptions) const
     {
-        EmitNewLinesAsNecessary(varInfo.GetOriginalLineNumber());
+        EmitEmptyLinesToLineNumber(varInfo.GetOriginalLineNumber());
 
         // from MSDN: https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx-graphics-hlsl-variable-syntax
         // [Storage_Class] [Type_Modifier] Type Name[Index] [: Semantic] [: Packoffset] [: Register]; [Annotations] [= Initial_Value]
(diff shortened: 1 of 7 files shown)


src/AzslcEmitter.cpp
```diff
@@ -488,6 +488,8 @@ namespace AZ::ShaderCompiler
 
     void CodeEmitter::EmitStruct(const ClassInfo& classInfo, string_view structName, const Options& options)
     {
+        EmitNewLinesAsNecessary(classInfo.GetOriginalLineNumber());
```
jeremyong-az: Suggestion: `PadLinesToLineNumber(classInfo.GetOriginalLineNumber())`

src/NewLineCounterStream.h
```diff
@@ -0,0 +1,80 @@
...
+    //! Wraps and forwards data to a std::ostream,
+    //! Overrides operator<< for char, const char * and string data and counts
+    //! the number of '\n' (new line) characters that go through it.
+    class NewLineCounterStream
+    {
+        typedef NewLineCounterStream Self;
+
+    public:
+
+        NewLineCounterStream(std::ostream& streamToWrap) : m_wrappedStream(streamToWrap)
```
jeremyong-az: Nit: `explicit`

src/NewLineCounterStream.h
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+            m_wrappedStream << str;
+            return *this;
+        }
+
+        template <class Any>
+        Self& operator<<(Any&& thing) // universal reference
+        {
+            m_wrappedStream << thing;
```
jeremyong-az: `std::forward<Any>(thing)` may avoid a copy if `thing` happened to be moved into this operator

src/NewLineCounterStream.h
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    m_lineCount++;
+                }
+            }
+            m_wrappedStream << str;
+            return *this;
+        }
+
+        template <class Any>
+        Self& operator<<(Any&& thing) // universal reference
```
jeremyong-az: Probably could drop the comment as this is a common/fundamental C++ idiom

src/NewLineCounterStream.h
```diff
@@ -0,0 +1,80 @@
...
+                    m_lineCount++;
+                }
+            }
+            m_wrappedStream << nts;
+            return *this;
+        }
+
+        Self& operator<<(const std::string& str)
+        {
+            for (auto it = str.begin(); it != str.end(); ++it )
```
jeremyong-az: Consider just `for (char c : str)` for conciseness

src/AzslcEmitter.cpp
galibzon: I'll change it to "EmitEmptyLinesToLineNumber".
In the class CodeEmitter, all functions that write to the output file start with Emit...

src/NewLineCounterStream.h
galibzon: Fixed.

src/NewLineCounterStream.h
galibzon: Comment removed here and from the original copy/pasted class 'DiagnosticStream'.

src/NewLineCounterStream.h
galibzon: I don't think that's right. std::forward only helps to forward as-is (lvalue as lvalue, rvalue as rvalue_, and mostly only helps to avoid compiler errors, which in this case is irrelevant because 'ostream<<' treats 'thing' as const &.

Granted, I don't consider myself a C++ lawyer, so if you have an example to show me where I'm wrong I'd appreciate it.

src/NewLineCounterStream.h
galibzon: Indeed. Done.

src/NewLineCounterStream.h
jeremyong-az: Not quite. Within the scope of `operator<<` here, `thing` is actually an _lvalue_ (not an rvalue). This is perhaps a common misconception, but the `&&` argument specifier is relevant only for ADL-binding to rvalues at the _call site_. Internally, the purpose of `std::forward` is to cast the argument to whatever its value-category was prior to entering the function body. Forwarding a universal reference is a pretty common idiom as a result, unless you know you'd rather pass an l-value (as is the case here).

If I have time later I'll try to provide a code snippet

galibzon: I did change the code to use std::forward, mostly because it's the "right" thing to do.

But I did many code snippets to get to trigger copies (without using std::forward) and it never happened. All I got, in some circumstances is compiler errors that disappeared using std::forward.

So I'd really appreciate a code snippet (which got to be obscure rather than common).

jeremyong-az (approved):

moudgils (approved):

PR #11, opened 2021-09-30
AZSLc Crash On Shader Option With Undefined Type

galibzon:
This is version 1.7.30

A clear error is now emitted when a shader option is of undefined type.

Previously it was only asserting and the user was NOT provided with
a message describing the root cause of the problem.

### DIFF
--- a/src/AzslcSemanticOrchestrator.cpp
+++ b/src/AzslcSemanticOrchestrator.cpp
@@ -570,7 +570,7 @@ namespace AZ::ShaderCompiler
             if (!varInfo.m_typeInfoExt.IsClassFound())
             {
                 const string message = FormatString("Unknown type '%s' for shader option '%.*s'",
-                    varInfo.m_typeInfoExt.GetDisplayName().c_str(), uqNameView.size(), uqNameView.data());
+                    varInfo.m_typeInfoExt.GetDisplayName().c_str(), static_cast<int>(uqNameView.size()), uqNameView.data());
                 ThrowAzslcOrchestratorException(ORCHESTRATOR_UNKNOWN_OPTION_TYPE, ctx->start, message);
             }
             // we'll set that here, an option is better flagged as static const for simplicity during emission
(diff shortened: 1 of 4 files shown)


jeremyong-az (approved):

moudgils (approved):

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -567,6 +567,12 @@ namespace AZ::ShaderCompiler
 
         if (isOption)
         {
+            if (!varInfo.m_typeInfoExt.IsClassFound())
+            {
+                const string message = FormatString("Unknown type '%s' for shader option '%s'",
+                    varInfo.m_typeInfoExt.GetDisplayName().c_str(), uqNameView.data());
```
santorac: If uqNameView is a string_view, you need to use '%.*s' and specify the length as an argument.

santorac: (if it happens to be a string, use c_str() instead of data())

santorac (changes requested):

src/AzslcSemanticOrchestrator.cpp
galibzon: uqNameView is of type UnqualifiedNameView, which is a subclass of 'string_view'.
Although 'string_view' "can be evil", in particular when you don't know where it comes from, e.g: A library, etc.
It is safe to use "UnqualifiedNameView.data()" throughout AZSLc  because they are initialized always from null terminated strings.

galibzon: To add more context...

UnqualifiedNameView is used heavily throughout AZSLc, and because of that you'll see many uses of ".data()".
Your request for change, to make an impact, would have to be applied throughout the code base.

And I'd be more than happy to do as many changes as necessary If I knew there were string_views in the code that came from somewhere else.

santorac (approved):

PR #12, opened 2021-10-10
Replace docopt with CLI11

jeremyong:
Replacing docopt with a non-regex based CLI parser removes the entirety
of the Boost dependency, which deflates to over 1.5 GB of source and
binary data (no longer needed after this change).

The CLI parser that replaces docopt generates a similar docstring and has
same parsing behavior. See https://github.com/CLIUtils/CLI11 to inspect CLI11
and its license (BSD 3).


jeremyong:
Looks like the tests complete 2 minutes faster relative to the last run also

galibzon (approved):
Major prophylaxis to AzslcMain.cpp. Awesome work. Thanks a lot!

moudgils (approved):

PR #13, opened 2021-10-11
non-initialized static const of any type seems to get a spurious 0

galibzon:
non-initialized static const of any type seems to get a spurious 0 initializer

This is version 1.7.32

Uninitialized static const variables now remain uninitilized.
The initialization value will be deferred to DXC which will
pick an appropriate value per HLSL spec.

Adjusted test cases.


jeremyong-az:
Just curious but is there any material change? I thought that static storage segment data in dxil binaries (like C++) are zero-initialized (even without an explicit initializer)

galibzon:
> 
> 
> Just curious but is there any material change? I thought that static storage segment data in dxil binaries (like C++) are zero-initialized (even without an explicit initializer)

For whatever reason the code was always assigning '= 0;' no matter what was the data type of the 'static const'.

jeremyong-az (approved):

santorac (approved):

PR #14, opened 2021-10-13
Changing inlineConstant to rootConstant keyword work.

galibzon:
This is version 1.7.33

Changed the text of the emitted reflection data for RootContants,
previously known as InlineConstants.


src/AzslcReflection.cpp
```diff
@@ -819,7 +819,7 @@ namespace AZ::ShaderCompiler
             uint32_t strideSize = BuildUserDefinedMemberLayout(structLayout, m_ir->m_rootConstantStructUID, options, options.m_packConstantBuffers, startAt, "");
             strideSize = GetPlatformEmitter().AlignRootConstants(strideSize);
 
-            rootConstantLayout["inputsForInlineConstants"] = structLayout;
+            rootConstantLayout["inputsForRootConstants"] = structLayout;
```
jeremyong-az, line 822: While we're here, is there any reason for the `ForRootConstants` suffix given that all these fields are embedded in the `RootConstantBuffer` object?

galibzon: I think the only reason for that naming style is that it facilitates searches.

an analogy is the hassle of searching for SystemComponent.cpp, which almost all Gems have one of those with the same file.

So, yes, technically speaking it is redundant, but it can help a lot when grepping.

jeremyong-az (approved):

moudgils (approved):

PR #16, opened 2021-10-15
Simplify antlr4 runtime integration

jeremyong:
- Adds o3de/antlr4 as a generation time dependency
- Removes unneeded python script to fetch git repos
- Adds antlr4_static as a target dependency
- Streamlines cmake options needed to ensure MSVC link compatibility


jeremyong:
Note, after pulling this change down, `src/external/antlr4` should be removed

jeremyong:
The tests pass locally on Windows, so I'll have to investigate why the linux CI is complaining (edit: fixed)

galibzon (approved):

moudgils (approved):

PR #17, opened 2021-10-17
Add preliminary [indirect] SRG Semantic attribute support

jeremyong:
This PR adds a new CLI option `--indirect` which if passed, activates any usages of the new `[indirect]` attribute on SRG Semantic declarations. Here's a code snippet to illustrate the behavior:

```hlsl
[indirect]
ShaderResourceGroupSemantic BindingPerObject
{
    FrequencyId = 0;
};

[indirect]
ShaderResourceGroupSemantic BindingPerMaterial
{
    FrequencyId = 1;
};

ShaderResourceGroup PerObject : BindingPerObject
{
    row_major float3x4 m_modelToWorld;
}

ShaderResourceGroup MaterialSrg : BindingPerMaterial
{
    float3 m_baseColor;
    float m_baseColorFactor;

    Texture2D m_baseColorMap;

    Sampler m_sampler
    {
       AddressU = Clamp;
       AddressV = Clamp;
       MinFilter = Linear;
       MagFilter = Linear;
       MipFilter = Linear;
    };
}

float4 MainPS(float2 uv : TEXCOORD) : SV_Target0
{
    float4 baseColorMapSample = MaterialSrg::m_baseColorMap.Sample(MaterialSrg::m_sampler, uv);
    float3 baseColor = baseColorMapSample.rgb * MaterialSrg::m_baseColor.rgb * MaterialSrg::m_baseColorFactor;

    return float4(baseColor, 1.0);
}
```

The SRG semantics above have `[indirect]` attributes, indicating that the shader author would like all UAV/SRV descriptors including the implicit struct to be accessed through a layer of indirection (to support ray tracing or indirect/instanced draws). After compiling the above with the flags `--use-spaces --indirect --entry-points=MainPS`, `azslc` will emit the following shader:

```hlsl

ByteAddressBuffer ManagedHeap_ByteAddressBuffer[] : register(t0, space100);
Texture2D ManagedHeap_Texture2D[] : register(t0, space101);
Texture3D ManagedHeap_Texture3D[] : register(t0, space102);
TextureCube ManagedHeap_TextureCube[] : register(t0, space103);
SamplerState ManagedHeap_SamplerState[] : register(s0, space104);
struct ManagedIndirection { uint index; uint offset; };

/* Generated code from  ShaderResourceGroup PerObject*/
struct PerObject_SRGConstantsStruct
{
    row_major float3x4 PerObject_m_modelToWorld;
};

ConstantBuffer<ManagedIndirection> PerObject_indirect : register(b0, space0);
static uint PerObject_base;
void setPerObject_base(uint instanceId) {
        PerObject_base = 8 * instanceId + PerObject_indirect.offset;
}

static ::PerObject_SRGConstantsStruct PerObject_SRGConstantBuffer;
void setPerObject_SRGConstantBuffer() {
        uint2 indexOffset = ManagedHeap_ByteAddressBuffer[PerObject_indirect.index].Load<uint2>(PerObject_base);
        PerObject_SRGConstantBuffer = ManagedHeap_ByteAddressBuffer[indexOffset.x].Load<::PerObject_SRGConstantsStruct>(indexOffset.y);
}

/* Generated code from  ShaderResourceGroup MaterialSrg*/
struct MaterialSrg_SRGConstantsStruct
{
     float3 MaterialSrg_m_baseColor;
     float MaterialSrg_m_baseColorFactor;
};

ConstantBuffer<ManagedIndirection> MaterialSrg_indirect : register(b0, space1);
static uint MaterialSrg_base;
void setMaterialSrg_base(uint instanceId) {
        MaterialSrg_base = 16 * instanceId + MaterialSrg_indirect.offset;
}

static ::MaterialSrg_SRGConstantsStruct MaterialSrg_SRGConstantBuffer;
void setMaterialSrg_SRGConstantBuffer() {
        uint2 indexOffset = ManagedHeap_ByteAddressBuffer[MaterialSrg_indirect.index].Load<uint2>(MaterialSrg_base);
        MaterialSrg_SRGConstantBuffer = ManagedHeap_ByteAddressBuffer[indexOffset.x].Load<::MaterialSrg_SRGConstantsStruct>(indexOffset.y);
}

static Texture2D MaterialSrg_m_baseColorMap;
void setMaterialSrg_m_baseColorMap() {
        MaterialSrg_m_baseColorMap = ManagedHeap_Texture2D[ManagedHeap_ByteAddressBuffer[MaterialSrg_indirect.index].Load<uint>(MaterialSrg_base + 8)];
}
static SamplerState MaterialSrg_m_sampler;
void setMaterialSrg_m_sampler() {
        MaterialSrg_m_sampler = ManagedHeap_SamplerState[ManagedHeap_ByteAddressBuffer[MaterialSrg_indirect.index].Load<uint>(MaterialSrg_base + 12)];
}

float4 MainPS(  float2 uv :TEXCOORD, uint instanceId :SV_InstanceID) :SV_Target0{
setPerObject_base(instanceId);
setPerObject_SRGConstantBuffer();
setMaterialSrg_base(instanceId);
setMaterialSrg_SRGConstantBuffer();
setMaterialSrg_m_baseColorMap();
setMaterialSrg_m_sampler();
float4 baseColorMapSample = ::MaterialSrg_m_baseColorMap . Sample ( ::MaterialSrg_m_sampler , uv ) ;
float3 baseColor = baseColorMapSample . rgb * ::MaterialSrg_SRGConstantBuffer.MaterialSrg_m_baseColor . rgb * ::MaterialSrg_SRGConstantBuffer.MaterialSrg_m_baseColorFactor ;
return float4 ( baseColor , 1.0 ) ;
}
```

Note the addition of the instance id, as well as setters which are invoked for every SRV/UAV. The stride and offsets are computed statically based on how many SRV/UAVs are present in an SRG, and are also reflected in the SRG and root signature. Note also that the new `--entry-points` CLI option is needed to emit the setters at the start of each entry point function. Another option would be to emit a getter per SRV/UAV, however, this generates significantly more bytecode, and may not always be optimized by the compiler.


jeremyong:
Aside from fixing some of the `AsError` tests that are broken, there are a still remaining items that need to be finished:

- [x] Fix test regressions
- [ ] Restrict use of CBs/StructuredBuffers in indirect SRGs
- [ ] Restricting indirect SRGs from possessing unbounded arrays
- [ ] Handling SRV/UAV descriptor fixed-size arrays
- [ ] Handling templated structured buffers
- [ ] Assigning register spaces for missing resource types (e.g. `Texture2D<uint>`, `Buffer<float4>`, etc.)

src/AzslcBackend.cpp
```diff
@@ -610,6 +610,45 @@ namespace AZ::ShaderCompiler
             RootSigDesc::SrgDesc srgDesc;
             srgDesc.m_uid = srgUid;
 
+            if (srgInfo->m_indirect && options.m_indirect)
+            {
+                // SRGs attached to an indirect SRG semantic do not bind resources directly into the root signature.
+                // Instead they index into the managed resource heaps with a secondary indirection. The indirection
+                // scheme is as follows (pseudo code):
+                //
+                //     ByteAddressBuffer ManagedHeap_ByteAddressBuffer[] : register(b0, space100);
```
moudgils, line 619: Suggestion: It would be good to add an example layout of this buffer to help understand how we are indexing into it.

src/AzslcBackend.cpp
```diff
@@ -610,6 +610,45 @@ namespace AZ::ShaderCompiler
...
+                //         &ManagedHeap_ByteAddressBuffer[ManagedIndirection.index].Load<uint>(
+                //             indirectionStride * SV_InstanceID + ManagedIndirection.offset);
+                //
+                // The data that starts at indirectionBase is a sequence of indirection constants needed to retrieve
+                // all the SRVs in the SRG, including the implicit struct data (which requires both a byte address buffer
+                // index and offset).
+
+                RootSigDesc::SrgParamDesc indirectParam = ReflectOneExternalResourceAndWrapWithUnifyIndices(srgInfo->m_indirectUID, bindInfo, rootSig);
+                indirectParam.m_type = RootParamType::SrgConstantCB;
+                indirectParam.m_num32BitConstants = 2;
```
moudgils, line 635: Add a quick comment as to why this is 2.

src/AzslcEmitter.cpp
```diff
@@ -959,29 +1020,128 @@ namespace AZ::ShaderCompiler
...
 
+    void CodeEmitter::EmitManagedResourceHeaps(const Options& options, const RootSigDesc& rootSig)
+    {
+        if (!options.m_indirect)
+        {
+            return;
+        }
+
+        // Emit unbounded arrays for each resource type
+        m_out
```
moudgils, line 1031: Do we need to bind/emit all the resource types even though the shader may only use some of them or none at all?

src/AzslcEmitter.cpp
```diff
@@ -959,29 +1020,128 @@ namespace AZ::ShaderCompiler
...
+        {
+            return;
+        }
+
+        std::string srgName = srgInfo.m_declNode->Name->getText();
+        std::string indirectHeapIndexOffset = srgName + "_indirect";
+        uint32_t indirectStride = static_cast<uint32_t>(srgInfo.m_srViews.size() + srgInfo.m_samplers.size()) * 4;
+        if (!srgInfo.m_implicitStruct.GetMemberFields().empty())
+        {
+            indirectStride += 8;
```
moudgils, line 1052: Needs a comment as to why this is 8

src/AzslcEmitter.cpp
```diff
@@ -959,29 +1020,128 @@ namespace AZ::ShaderCompiler
...
+            << "Texture2D ManagedHeap_Texture2D[] : register(t0, space101);\n"
+            << "Texture3D ManagedHeap_Texture3D[] : register(t0, space102);\n"
+            << "TextureCube ManagedHeap_TextureCube[] : register(t0, space103);\n"
+            << "SamplerState ManagedHeap_SamplerState[] : register(s0, space104);\n"
+            << "struct ManagedIndirection { uint index; uint offset; };\n";
+    }
+
+    void CodeEmitter::EmitSRGIndirectionSetters(const SRGInfo& srgInfo, const IdentifierUID& srgId, const Options& options, const RootSigDesc& rootSig)
+    {
+        if (!options.m_indirect)
```
moudgils, line 1042: It would be good to add the comment with the example in the PR description here.

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -1468,6 +1486,57 @@ namespace AZ::ShaderCompiler
...
+        // Establish a canonical ordering for all SRV indirection constants
+        uint32_t offset = 0;
+
+        if (!srgInfo.m_implicitStruct.GetMemberFields().empty())
+        {
+            // If there are implicit members, always assign the initial two uints to the implicit struct index/offset pair
+            offset += 2;
+        }
+
+        for (const auto& id : srgInfo.m_srViews)
```
moudgils, line 1523: should m_srViews be m_srgViews?

src/AzslcEmitter.cpp
jeremyong-az: yup it's 2 uints so 8 bytes, but I need to do a pass to document all of this better

src/AzslcSemanticOrchestrator.cpp
jeremyong-az: No this SR stands for "shader resource" so the entire thing reads "shader resource views" aka SRVs. I would have authored this as `m_srvs` personally, but this is just what's in the codebase currently

src/AzslcEmitter.cpp
jeremyong-az: They get stripped out in the final dxil thankfully (did confirm this), but in reality there's no actual cost because they all will ultimately map to the same descriptor range

santorac:
I suggest making the "indirect" be like a member of the SRG Semantic definition rather than an attribute. The original design for SRG Semantic is that you do all your configuration inside it by setting something that looks like variables. There's a variable called FrequencyId, one called FallbackKeyBitSize (or something like that), and you would add one called "Indirect=true".

moudgils:
In general I like this approach. Add SuperVariant system to this and you can fallback to the direct (current) approach easily. 

On the c++ side in order to get this up and running quickly all the ManagedHeaps can be different SRGs (SRGs currently already support unbounded image and buffer views) which are bound only once in a frame. All the current direct SRGs will have a counter indirect srg (for index/offset) and depending on which pathway we follow based on the hw we bind that srg. Maybe we extend the SRG at the RHI level to help with this as oppose to forcing the higher level code to manage two differnt SRGs?

jeremyong:
I think in practice there will be modifying CLI options to control how the managed heaps are bound and also how the indirection constants are pulled. Agreed that there is probably a shorter path to getting this to work based on how the SRG system works currently. I need to do more code reading to understand the SRG compilation deeper to make any decisions there.

src/AzslcEmitter.cpp
```diff
@@ -752,7 +783,19 @@ namespace AZ::ShaderCompiler
         // emit parameters:
         m_out << "(";
         bool inhibitInitializers = AlreadyEmittedFunctionDeclaration(uid) && !emitAsDeclaration;
-        EmitParameters(funcSub.GetParameters(emitAsDeclaration).begin(), funcSub.GetParameters(emitAsDeclaration).end(), options, !inhibitInitializers);
+        auto& parameters = funcSub.GetParameters(emitAsDeclaration);
+        EmitParameters(parameters.begin(), parameters.end(), options, !inhibitInitializers);
+
+        if (!preamble.empty())
```
siliconvoodoo, line 789: To cleanup this uncanonical addendum, I'd go in IntermediateRepresentation::MiddleEnd method, and add the parameter injection as a proper variable in the IR. Not sure what it could break but it seems cleaner.

src/AzslcEmitter.cpp
```diff
@@ -959,29 +1020,128 @@ namespace AZ::ShaderCompiler
...
+        }
+
+        // Emit unbounded arrays for each resource type
+        m_out
+            << "ByteAddressBuffer ManagedHeap_ByteAddressBuffer[] : register(t0, space100);\n"
+            << "Texture2D ManagedHeap_Texture2D[] : register(t0, space101);\n"
+            << "Texture3D ManagedHeap_Texture3D[] : register(t0, space102);\n"
+            << "TextureCube ManagedHeap_TextureCube[] : register(t0, space103);\n"
+            << "SamplerState ManagedHeap_SamplerState[] : register(s0, space104);\n"
+            << "struct ManagedIndirection { uint index; uint offset; };\n";
```
siliconvoodoo, line 1037: these variables are probably better off inserted in the IR in IntermediateRepresentation::MiddleEnd too. This way symbol collision can't happen, and the Emitter will automatically do this. Could be worth it investigating if that harcoded "100" shouldn't be configured explicitely in the ShaderResourceGroupSemantic, alongside `FrequencyId=1`, you'd have `ManagedHeapSpaceBase=100` or something.

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -1468,6 +1486,57 @@ namespace AZ::ShaderCompiler
...
+        bool indirect = get<SRGSemanticInfo>(srgSemanticClassInfo.m_subInfo).m_indirect;
+        srgInfo.m_indirect = indirect;
+
+        if (!srgInfo.m_indirect)
+        {
+            return;
+        }
+
+        std::string indirectUIDName = GetCurrentScopeIdAndKind().first.m_name + "_indirect";
+        srgInfo.m_indirectUID = IdentifierUID{ QualifiedName{ indirectUIDName } };
```
siliconvoodoo, line 1509: don't construc the ID yourself, it will be given after you AddIdentifier

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -1468,6 +1486,57 @@ namespace AZ::ShaderCompiler
...
+        const ClassInfo& srgSemanticClassInfo = srgSemanticSym->second.GetSubRefAs<ClassInfo>();
+        bool indirect = get<SRGSemanticInfo>(srgSemanticClassInfo.m_subInfo).m_indirect;
+        srgInfo.m_indirect = indirect;
+
+        if (!srgInfo.m_indirect)
+        {
+            return;
+        }
+
+        std::string indirectUIDName = GetCurrentScopeIdAndKind().first.m_name + "_indirect";
```
siliconvoodoo, line 1508: don't store these stuff under raw strings, an identifer is either qualified or unqualified. the scope ID is qualified, so you're constructing a QualifiedName.

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -1468,6 +1486,57 @@ namespace AZ::ShaderCompiler
...
+        srgInfo.m_indirectUID = IdentifierUID{ QualifiedName{ indirectUIDName } };
+        AddIdentifier(UnqualifiedNameView{ std::string_view{ indirectUIDName.c_str() + 1, indirectUIDName.size() - 1 } }, Kind::Variable)
```
siliconvoodoo, line 1510: This sort of mangling hacking is better done with The Mangling tools in AzslcMangling.h It seems you're trying to remove the leading `/`.
`AddIdentifier` takes an unqualified name because it adds in context (in the current scope).
You don't know how this will hold in the future if the SRG can be declared in namespaces. Instead, use `ExtractLeaf` directly on the scopeid.

jeremyong: Thanks for all the implementation advice! This is a fairly new codebase for me but it definitely needs a cleanup pass with a number of items suggested

src/AzslcEmitter.cpp
jeremyong: Agreed this is a weekend prototype and the spaces reserved for these heaps needs to be specifiable with a CLI option (plus default) and reflected back in the emitted layout.

PR #18, opened 2021-10-20
Atom/galibzon/pad to n

galibzon:
Add [[pad_to(N)]] language feature 

This is version 1.7.34

"struct" definitions now support the [[pad_to(N)]] attribute.
This attribute, when used, must be declared after a member variable.
Each member variable can have one of these attributes after it.

The attribute is useful to pad the data with dummy variables until
guaranteeing that the offset of the next variable starts with the
desired alignment. The 'N' argument is a single integral literal
that must be a multiple of 4.

A companion class to the IntermediateRepresentation has been added:
StructPadToMutator. It contains all the logic and algorithms to
collect, during parsing, the variables that require padding
and apply the appropriate mutations (variable insertions) during
the IntermediateRepresentation::MiddleEnd().


src/AzslcIntermediateRepresentation.cpp
```diff
@@ -92,8 +94,13 @@ namespace AZ::ShaderCompiler
                     m_metaData.m_outputFormatHint[rtIndex] = hint;
                 }
             }
+        } else if (attrInfo.m_attribute == "pad_to")
```
jeremyong-az: Nit: `\n`

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+            m_ir.ThrowAzslcIrException(IR_INVALID_PAD_TO_ARGUMENTS, attrInfo.m_lineNumber, errorMsg);
+        }
+        //Is the argument integral?
+        if (!holds_alternative<ConstNumericVal>(attrInfo.m_argList[0]))
+        {
+            string errorMsg("The [[pad_to(N)]] attribute only accepts one argument of integral type. A non integral argument was given.");
+            m_ir.ThrowAzslcIrException(IR_INVALID_PAD_TO_ARGUMENTS, attrInfo.m_lineNumber, errorMsg);
+        }
+        // Read the integral.
+        auto pad_to_value = static_cast<uint32_t>(ExtractValueAsInt64(get<ConstNumericVal>(attrInfo.m_argList[0]), 0));
```
jeremyong-az: Might be worth adding `ExtractValueAsUint32` since this is probably a common case

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+        if (!pad_to_value)
+        {
+            string errorMsg("Failed to read input integral to [[pad_to(N)]].");
+            m_ir.ThrowAzslcIrException(IR_INVALID_PAD_TO_ARGUMENTS, attrInfo.m_lineNumber, errorMsg);
+        }
+        // Must be a multiple of 16
```
jeremyong-az: "multiple of 4"

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        auto& [curScopeId, curScopeKindInfo] = m_ir.GetCurrentScopeIdAndKind();
+        if (curScopeKindInfo.IsKindOneOf(Kind::Struct))
+        {
+            // We need to get the variable declared before this attribute.
+            auto varUid = m_ir.GetLastMemberVariable(curScopeId);
+            if (varUid.IsEmpty())
+            {
+                auto errorMsg = FormatString("The [[pad_to(N)]] attribute must be added after a member variable inside 'struct' or 'class'. The current scope '%.*s' doesn't have a declared variable yet.",
```
jeremyong-az: Should we plan on permitting this in the future? (thinking of partial structs)

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+                static_cast<int>(curScopeId.GetName().size()), curScopeId.GetName().data());
```
jeremyong-az: Nit: spacing

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+            {
+                const auto rows = exportedType.m_arithmeticInfo.m_rows;
+                const auto cols = exportedType.m_arithmeticInfo.m_cols;
+                const auto packAlignment = exportedType.m_arithmeticInfo.IsMatrix() ? Packing::Alignment::asMatrixStart : Packing::Alignment::asVectorStart;
+                startAt = offset = Packing::AlignOffset(layoutPacking, offset, packAlignment, rows, cols);
+            }
+
+            uint32_t totalArraySize = 1;
+            ArrayDimensions listOfArrayDim = varInfo.m_typeInfoExt.GetDimensions();
+            std::reverse(listOfArrayDim.m_dimensions.begin(), listOfArrayDim.m_dimensions.end());
```
jeremyong-az: Nit: might be better to use a reverse iterator instead of allocating in the `ArrayDimensions` internal vector and copying

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+        const auto* classInfo = m_ir.GetSymbolSubAs<ClassInfo>(exportedTypeId.m_name);
+        for (const auto& memberField : classInfo->GetMemberFields())
+        {
+            const auto currentStride = tempOffset;
+            CalculateMemberLayout(memberField, false, emitRowMajors, GetExtendedLayout(layoutPacking), tempOffset);
+            largestMemberSize = std::max(largestMemberSize, tempOffset - currentStride);
+        }
+
+        tempOffset = Packing::AlignStructToLargestMember(layoutPacking, tempOffset, largestMemberSize);
```
jeremyong-az: Quick question: how exactly is the largest member impacting the overall alignment? Could you provide a quick example of what's happening here?

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,479 @@
...
+            {
+                newVarInfo.m_typeInfoExt = ExtendedTypeInfo{ m_ir.m_sema.CreateTypeRefInfo(padType),
+                    {}, {{itemsCount}}, {}, Packing::MatrixMajor::Default };
+            }
+            newVarKind.GetSubRefAs<VarInfo>() = newVarInfo;
+            return newVarUid;
+        };
+
+        // The key idea is to add, at most, three variables. They will be added depending on keeping a 16-byte alignment from @startingOffset
+        // 1- The first variable will be added if @startingOffset is not 16-bytes aligned. It will be a float, float2 or float3.
```
jeremyong-az: Out of curiosity, in what cases is this first variable needed? I imagine that between just the `float4`s plus any needed trailing padding bits this would be covered (the first `float4` if needed would align to a `float4` boundary, and the trailing bits would take care of the rest)

src/AzslcIntermediateRepresentation.cpp
galibzon: Fixed.

src/StructPadToMutator.cpp
galibzon: Good idea. Added a template version called ExtractValueAs<T>()

src/StructPadToMutator.cpp
galibzon: ooops. Fixed.

src/StructPadToMutator.cpp
galibzon: When the time comes and the need arises, it will be "relatively" easy to add.

src/StructPadToMutator.cpp
galibzon: Fixed.

src/StructPadToMutator.cpp
galibzon: CalculateMemberLayout() & CalculateUserDefinedMemberLayout() are verbatim copies of
CodeReflection::BuildMemberLayout() & CodeReflection::BuildUserDefinedMemberLayout(), except that I removed the JSON code from it.

src/StructPadToMutator.cpp
galibzon: Not sure, see my comment above. Those functions are varbatim copies of existing code that is used to calculate the SRG layout data, etc.

src/StructPadToMutator.cpp
galibzon: For non-ConstantBuffer packing the float4 is not automatically aligned to 16 bytes.
Example:
```
struct MyStructA
{
         float m_data;
	 float4 m_arr[2];
};
```
For ConstantBuffer case you'll get these offsets:
```
;                   float m_data;                     ; Offset:    0
;                   float4 m_arr[2];                  ; Offset:   16
```
For StructuredBuffer case you'll get:
```
;           float m_data;                             ; Offset:    0
;           float4 m_arr[2];                          ; Offset:    4
```

src/StructPadToMutator.h
```diff
@@ -0,0 +1,117 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ * 
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+/*
+#include "AzslcUtils.h"
+*/
```
santorac: Delete this?

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+        auto& [curScopeId, curScopeKindInfo] = m_ir.GetCurrentScopeIdAndKind();
+        if (curScopeKindInfo.IsKindOneOf(Kind::Struct))
+        {
+            // We need to get the variable declared before this attribute.
+            auto varUid = m_ir.GetLastMemberVariable(curScopeId);
+            if (varUid.IsEmpty())
+            {
+                auto errorMsg = FormatString("The [[pad_to(N)]] attribute must be added after a member variable inside 'struct' or 'class'."
+                                             " The current scope '%.*s' doesn't have a declared variable yet.",
```
santorac: Is this right?
```suggestion
                auto errorMsg = FormatString("The [[pad_to(N)]] attribute must be added after a member variable inside 'struct'."
                                             " The current scope '%.*s' doesn't have a declared variable yet.",
```

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+            if (structItor == m_structsToPad.end())
+            {
+                m_structsToPad[curScopeId] = MapOfVarInfoUidToPadding();
+            }
+            MapOfVarInfoUidToPadding& varInfoUidToPadMap = m_structsToPad[curScopeId];
+            auto varInfoItor = varInfoUidToPadMap.find(varUid);
+            if (varInfoItor != varInfoUidToPadMap.end())
+            {
+                // It appears that there are two consecutive [[pad_to(N)]] attributes. This is an error.
+                auto errorMsg = string("Two consecutive [[pad_to(N)]] attributes are not allowed inside 'struct' or 'class'");
```
santorac:
```suggestion
                auto errorMsg = string("Two consecutive [[pad_to(N)]] attributes are not allowed inside 'struct'");
```

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+        if (visitedStructs.count(structUid))
+        {
+            return;
+        }
+        
+        const auto* classInfoPtr = m_ir.GetSymbolSubAs<ClassInfo>(structUid.m_name);
+        const auto listOfPairs = GetVariablesOfStructTypeThatRequirePadding(classInfoPtr);
+        for (const auto& [typeUid, varUid] : listOfPairs)
+        {
+            StructUidSortVisitFunction(typeUid, visitedStructs, sortedList);
```
santorac: Is it possible to define structs that reference each other and still get to this point in the code? If so, you might hit infinite recursion and crash.

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+        }
+        
+        visitedStructs.insert(structUid);
+        sortedList.push_back(structUid);
+    }
+
+
+    vector<pair<IdentifierUID, IdentifierUID>> StructPadToMutator::GetVariablesOfStructTypeThatRequirePadding(const ClassInfo* classInfo) const
```
santorac: I'm curious why is padding supported for struct and not for class?

src/AzslcUtils.h
```diff
@@ -230,6 +230,36 @@ namespace AZ::ShaderCompiler
...
+            {
+                return *defval;
+            }
+        }
+        else if (holds_alternative<float>(var))
+        {
+            // Casting from float has precision loss: https://onlinegdb.com/By0AJTUVE
+            auto floatVal = get<float>(var);
+            auto intVal = static_cast<int64_t>(floatVal);
+            PrintWarning(Warn::W3, none, "warning: Casting float ", floatVal, " to integer, will result in ", intVal);
```
jeremyong-az, line 252: This warning only applies if `T` is an integral type. We may need a small pass on this function to account for the various types `T` could vary over

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+                {
+                    if (!m_ir.GetIdAndKindInfo(varInfo.GetTypeId().m_name))
+                    {
+                        continue;
+                    }
+
+                    // If array is a structure
+                    if (IsProductType(varClass))
+                    {
+                        size = CalculateUserDefinedMemberLayout(exportedType.m_typeId, emitRowMajor, layoutPacking, startAt = offset);
```
santorac: Using "startAt = offset" inline could lead to confusion. I suggest putting this on its own line and pass startAt instead.

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+            uint32_t totalArraySize = 1;
+            ArrayDimensions listOfArrayDim = varInfo.m_typeInfoExt.GetDimensions();
+            std::reverse(listOfArrayDim.m_dimensions.begin(), listOfArrayDim.m_dimensions.end());
+            for (const auto dim : varInfo.m_typeInfoExt.GetDimensions())
+            {
+                totalArraySize *= dim;
+            }
+
+            if (varInfo.m_typeInfoExt.IsArray() && !isArrayItr)
```
santorac: What is "isArrayItr"? Could this variable be named better?

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+                continue;
+            }
+
+            const auto bytesToAdd = alignedOffset - nextMemberOffset;
+            idx += InsertPaddingVariables(classInfo, structUid, idx+1, nextMemberOffset, bytesToAdd);
+            nextMemberOffset = alignedOffset;
+        }
+    }
+
+    uint32_t StructPadToMutator::CalculateMemberLayout(const IdentifierUID& memberId,
```
santorac: It's pretty hard to follow what's going on in this function. You might consider adding more comments. I know you have some already, but it wasn't enough for me to easily understand.

src/StructPadToMutator.cpp
```diff
@@ -0,0 +1,480 @@
...
+    {
+        // Alignment start
+        uint32_t tempOffset = startAt = Packing::AlignOffset(layoutPacking, startAt, Packing::Alignment::asStructStart, 0, 0);
+
+        uint32_t largestMemberSize = 0;
+
+        const auto* classInfo = m_ir.GetSymbolSubAs<ClassInfo>(exportedTypeId.m_name);
+        for (const auto& memberField : classInfo->GetMemberFields())
+        {
+            const auto currentStride = tempOffset;
```
santorac: Nitpick: only use auto when it's saving long and ugly typenames. Here the type is "uint32_t" so just put that to make the code more readable.

src/StructPadToMutator.cpp
santorac: Consider including this info in code comments, not just PR comments.

santorac (commented):
I didn't finish reviewing every file yet. I need to take a break and come back later.

src/StructPadToMutator.h
galibzon: Deleted. Thanks.

src/StructPadToMutator.cpp
galibzon: Fixed. Thanks.

src/StructPadToMutator.cpp
galibzon: I was thinking about that. Let me run some experiments. Maybe other earlier pieces of code prevent this from happening.

src/AzslcUtils.h
galibzon: Oh yeah. Lots of things to check on T.

src/StructPadToMutator.cpp
galibzon: Fixed.

src/StructPadToMutator.cpp
galibzon: I removed support for 'class' types at the last minute. It was working, the algorithm even works for SRG.
Until I tried StructuredBuffer<class type> and DXC doesn't like that at all. The most common usage of [[pad_to]] is to facilitate the case of StructureBuffers and tightly packed structs that can also be used in ConstantBuffers.

If there's value in supporting 'class'es or SRGs the support is trivial to add.

src/StructPadToMutator.cpp
galibzon: Good catch. That's copy/pasta from the original code that is used for SRG layout calculation. Fixed.

src/StructPadToMutator.cpp
galibzon: As I mentioned in other comments: 
CalculateMemberLayout() & CalculateUserDefinedMemberLayout() are verbatim copies of
CodeReflection::BuildMemberLayout() & CodeReflection::BuildUserDefinedMemberLayout(), except that I removed the JSON code from it.

Lots of ugly code in there.

isArrayItr is flag that says the data type is iterating over an array definition.

I only removed the JSON related code from the original because it's code that's been working fine for the last two years.

src/StructPadToMutator.cpp
galibzon: A neither do I understand much of it. Other than it's the same code used during SRG layout calculations.

src/StructPadToMutator.cpp
galibzon: Agree, I hate auto (for the most part). It's easy on the coder, but really bad for the reviewer.
But the code base of AZSLc changed my soul.

src/AzslcUtils.h
santorac: It looks like you only intend to use this for ints and floats. You could implement the function in cpp instead of h, and then explicitly instantiate just int and float. That will make it very clear what types are supported, while limiting the scope of error checking you need to do.

tests/Syntax/AsErrors/srg-pad-to.azsl
```diff
@@ -0,0 +1,19 @@
...
+
+ShaderResourceGroupSemantic slot1
+{
+    FrequencyId = 1;
+};
+
+ShaderResourceGroup DemoSrg : slot1
+{
+    float m_red;
+    [[pad_to(16)]]
```
santorac: Maybe add more error cases to cover each of the many error checks  you put in the code: wrong args for pad_to, pad_to before any members, two pad_two's in a row, etc.

tests/Advanced/struct-pad-to.azsl
```diff
@@ -0,0 +1,44 @@
...
+    [[pad_to(256)]]
+};
+
+ShaderResourceGroup DemoSrg : slot1
+{
+    MyStruct m_myStruct;
+    StructuredBuffer<MyStruct> m_mySB;
+};
+
+float4 PSMain() : SV_Target0
```
santorac: Does this test sufficiently cover all the variations of padding variables? You had the whole thee-variable "float1-3, float4[N], float1-3" thing going on in the code. I can't tell whether those cases are all covered here.

src/StructPadToMutator.cpp
galibzon: Did some validations and AZSLc fails earlier when there are recursive usages of struct.

tests/Syntax/AsErrors/srg-pad-to.azsl
galibzon: Done.

tests/Advanced/struct-pad-to.azsl
galibzon: Added new test case that pads to 252, and exercises the case where three variables must be added to complete the padding.

src/StructPadToMutator.cpp
jeremyong-az: There's definitely value. One of the uses IIRC was adding padding to an SRG to ensure a value was an aligned load on Android

santorac (approved):

src/StructPadToMutator.cpp
galibzon: Good to know, will add support for classes and SRGs.

galibzon: [[pad_to(N)]] is now supported for classes and SRGs.

jeremyong-az: Excellent!

src/StructPadToMutator.cpp
galibzon: Added the whole comment to the this function.

jeremyong-az (approved):

PR #23, opened 2022-02-04
Improved project generation and compilation for Linux.

galibzon:
In Linux, AZSLc can now be compiled with clang-12. By default it uses g++.

Optimized the git clone operation of ANTLR to do a shallow clone.

Added a helper script 'clean_build.sh' to clean the linux/mac build folders.

Signed-off-by: galibzon <garrieta@amazon.com>

### DIFF
--- a/clean_build.sh
+++ b/clean_build.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/bash
+
+DIR_LIST=("bin/" "build/" "src/dist/" " src/external/antlr4/")
+for DIR in ${DIR_LIST[@]}; do
+    echo "Removing $DIR..."
+    rm -rf $DIR
+    echo "Done."
+done
+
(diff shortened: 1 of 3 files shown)


santorac (approved):

jeremyong-az:
Is there a reason clang-12 specifically is required? When using gcc for example, no version is stipulated.

galibzon:
The only reason to use clang-12 is that it is the version being used when compiling O3DE for Linux. In ubuntu20.04 the  "clang" command defaults to clang-10. Granted, compiling azslc with clang-12 is just for exploratory purposes. gcc is still the default in case "clang" is not provided when executing "prepare_solution_linux.sh".
With regards to gcc, just kept it the way it was to use the default.

jeremyong-az (approved):

jeremyong-az:
OK, it may be worth revisiting this assumption in the future, since other distros don't necessarily alias the clang compilers the same way Ubuntu does (in fact, most don't) and we would want this project to be easily usable on other distros also.

PR #25, opened 2022-03-20
Support multiple unbounded arrays within a single SRG

jeremyong-az:
DX12 binding semantics assign a register to each binding in an array and
do not permit overlap. Spaces act as a "namespace" to permit multiple
unbounded arrays to be present in a shader. Instead of restricting
unbounded array use within an SRG, this change modifies the space
assignment strategy. Unbounded arrays (on platforms that require it,
DX12 in this case) are designated a unique space, starting from 1000 and
counting up. This "spill space" is orthogonal to the space assignment
corresponding with the SRG frequency.

### DIFF
--- a/src/AzslcBackend.cpp
+++ b/src/AzslcBackend.cpp
@@ -228,10 +228,17 @@ namespace AZ::ShaderCompiler
         }
     }
 
-    void MultiBindingLocationMaker::SignalRegisterIncrement(BindingType regType, int count /* = 1*/)
+    void MultiBindingLocationMaker::SignalRegisterIncrement(BindingType regType, int count, bool isUnbounded)
     {
         m_untainted.m_registerPos[regType] += count;
         m_merged.m_registerPos[regType] += count;
+
+        if (isUnbounded)
+        {
+            m_untainted.m_unboundedSpillSpace = m_unboundedSpillSpace;
+            m_merged.m_unboundedSpillSpace = m_unboundedSpillSpace;
+            ++m_unboundedSpillSpace;
+        }
     }
 
     BindingPair MultiBindingLocationMaker::GetCurrent(BindingType regType)
@@ -573,9 +580,19 @@ namespace AZ::ShaderCompiler
         }
 
         auto regType = RootParamTypeToBindingType(paramType);
-        auto srgElementDesc = RootSigDesc::SrgParamDesc{ id, paramType, bindInfo.GetCurrent(regType), count, -1, isUnboundedArray};
+
+        auto srgElementDesc = RootSigDesc::SrgParamDesc{
+            id,
+            paramType,
+            bindInfo.GetCurrent(regType),
+            count,
+            -1,
+            bindInfo.m_unboundedSpillSpace,
+            isUnboundedArray };
+
         rootSig.m_descriptorMap.emplace(id, srgElementDesc);
-        bindInfo.SignalRegisterIncrement(regType, count);
+        bindInfo.SignalRegisterIncrement(regType, count, isUnboundedArray);
+
         return srgElementDesc;
     }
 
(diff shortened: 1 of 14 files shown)


jeremyong-az:
@galibzon The test that's failing in actions passes for me locally, was curious if you had any ideas

image

jeremyong-az:
A change is needed to correct tests that expect an error with multiple unbounded arrays present.

moudgils (approved):

Adi-Amazon (approved):
Great enhancement.
Could we add in the documentation of the header several lines about this special case and explain the choice of using indirection into the the upper spaces (DX12) instead of within the Srg space?

galibzon (changes requested):
What test cases were added to validate these changes?
https://github.com/o3de/o3de-azslc/wiki/Adding-New-Test-Cases-To-AZSLc

jeremyong-az:
@galibzon None as of yet, although a file was added to act as the test once written. The thing that needs to be done first before this PR can be merged is fixing tests that were broken because limits that existed previously are now lifted.

galibzon:
@jeremyong-az , also it's important to highlight that azslc follows the exact same workflow as o3de. o3de/o3de-azslc is upstream, aws-lumberyard-dev/o3de-azslc is origin. So the PR should come from aws-lumberyard-dev/o3de-azslc.

jeremyong-az:
@galibzon I did not have permissions to aws-lumberyard-dev/o3de-azslc when this PR was created.

moudgils:
Let me move this pr to the correct repo, address the failing tests and maybe add a new one to test the new changes.

src/AzslcBackend.h
```diff
@@ -98,6 +101,10 @@ namespace AZ::ShaderCompiler
         int GetAccumulated(BindingType r) const;
 
         int m_space = 0;  //<! logical space
+
+        // See: MultiBindingLocationMarker::m_unboundedSpillSpace
+        int m_unboundedSpillSpace;
```
santorac, line 106: Let's initialize this just to be safe.

src/AzslcBackend.h
```diff
@@ -109,21 +116,28 @@ namespace AZ::ShaderCompiler
     class MultiBindingLocationMaker
     {
     public:
-        MultiBindingLocationMaker(const Options& options)
-            : m_options(options)
+        MultiBindingLocationMaker(const Options& options, int& unboundedSpillSpace)
+            : m_options{ options }
+            , m_unboundedSpillSpace{ unboundedSpillSpace }
```
santorac, line 121: I would prefer unboundedSpillSpace be a pointer instead of a reference, to make it obvious what's going on at the call site. Calling "MultiBindingLocationMaker(options, &unboundedSpillSpaceValue)" is a more clear than calling "MultiBindingLocationMaker(options, unboundedSpillSpaceValue)"

src/AzslcBackend.cpp
```diff
@@ -588,7 +605,10 @@ namespace AZ::ShaderCompiler
 
     RootSigDesc Backend::BuildSignatureDescription(const Options& options, int num32BitConst) const
     {
-        MultiBindingLocationMaker bindInfo{ options };
+        // We start at an arbitrarily high number to avoid conflicts with spaces occupied by any user declarations
+        m_unboundedSpillSpace = 1000;
```
santorac, line 609: I see that m_unboundedSpillSpace is a mutable member, so it can be assigned here in a const function. Why can't this just be a local variable instead? Is it used after the BuildSignatureDescription function closes? If so, please add a code comment about what's going on, as this is unexpected.

tests/Samples/UnboundedArrays2.azsl_manual
```diff
@@ -0,0 +1,24 @@
...
+{
+
+    Texture2D<float4>        m_texSRVa[];      // t0+, space1000
+    Texture2D<float4>        m_texSRVb[];      // t0+, space1001
+    RWTexture2D<float4>      m_texUAVa[];      // u0+, space1002
+    RWTexture2D<float4>      m_texUAVb[];      // u0+, space1003
+    Sampler                  m_samplera[];     // s0+, space1004
+    Sampler                  m_samplerb[];     // s0+, space1005
+    ConstantBuffer<MyStruct> m_structArraya[]; // b0+, space1006
+    ConstantBuffer<MyStruct> m_structArrayb[]; // b0+, space1007
```
santorac, line 23: Should this be an Emission test instead, so we can check the appropriate spaces are used in the generated output?

santorac (changes requested):

Adi-Amazon (commented):
Very good enhancement. Is there a fast course of action to move this to the global public repo?

PR #42, opened 2022-06-23
Added Support For Unlimited Unbounded Arrays

santorac:
This PR builds on top of a prototype by @jeremyong-az to introduce the concept of a register spill space. Every unbounded array will be assigned a unique register space starting at 1000. This is meant to support unlimited unbounded arrays on dx12, regardless of resource type or which SRG holds the resources. (dx12 is different from other platforms in that unbounded arrays consume the entire remaining register space for that resource type).

The original prototype only did this for DirectX but we realized it could be done the same for every platform, as DirectX is the only one that cares about the concept of register spaces. The other platforms just ignore the spaces anyway. So I was able to simplify the code by just doing everywhere. This also removes the need most of the unbounded arrays validation it was doing before, and several automated tests.

Testing:
Tested with bindless feature branches https://github.com/o3de/o3de/pull/8410 and https://github.com/o3de/o3de-atom-sampleviewer/pull/381. Used the AzslCompilerOverridePath registry setting to force the new azslc.exe. Made local changes to remove "use-spaces".
ASV RHI/BindlessPrototype works fine.
ASV Dx12 full test suite passes with the exception of a known issue. https://github.com/o3de/o3de-atom-sampleviewer/issues/464
**ASV Vulkan full test suite crashes in raytracing code.**


jeremyong-az (approved):
Nice!

moudgils (approved):

siliconvoodoo:
I believe all of this work has been integrated in development by @rgba16f ? we can close the PR?

PR #51, opened 2022-07-21
Allow proper parsing of unsigned long long literal ULL

martinwinter-huawei:
Signed-off-by: Martin Winter <martin.winter@huawei.com>

While `hlsl` supports 64 bit unsigned integer types (i.e. `uint64_t`), using the correct literal suffix `ULL` leads to a parsing error.
So code like this
```cpp
uint64_t test_variable = 1ULL << 35;
```
will not compile and the error message typically is
```
syntax error #1: missing ';' at 'L'
```
This commit updates the lexer definition to define `ULL` or `ull` as a valid `IntergerSuffix`.


martinwinter-huawei:
Superseded by #53

PR #52, opened 2022-07-22
Fix usage of `exportKeywords.py` by using the correct flags

martinwinter-huawei:
Signed-off-by: Martin Winter <martin.winter@huawei.com>

The `exportKeywords.py` script can be used to generate a list of predefined types. The current state did not have a variable set (`silent` was not set but `False` passed directly) and the argument parser did not properly receive the `--listpredefined` flag, hence always crashed.

### DIFF
--- a/src/AzslcMain.cpp
+++ b/src/AzslcMain.cpp
@@ -398,7 +398,7 @@ int main(int argc, const char* argv[])
     cli.add_flag("-r", visitRecursively, "(Option of --visitsym) Visit recursively.");
 
     bool listPredefined = false;
-    cli.add_flag("--listpredefined", "Output a list of all predefined types in AZSLang.");
+    cli.add_flag("--listpredefined", listPredefined, "Output a list of all predefined types in AZSLang.");
 
     int maxSpaces = std::numeric_limits<int>::max();
     auto maxSpacesOpt = cli.add_option("--max-spaces", maxSpaces, "Will choose register spaces that do not extend past this limit.");
--- a/tests/testfuncs.py
+++ b/tests/testfuncs.py
@@ -145,10 +145,9 @@ def buildAndGetError(thefile, compilerPath, silent, extraArgs):
 def dumpKeywords(compilerPath):
     '''returns tuple: (categorized-tokens-in-algebraic-form, ok-bool)'''
     import clr
-    silent = False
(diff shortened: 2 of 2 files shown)


tests/testfuncs.py
```diff
@@ -145,7 +145,8 @@ def buildAndGetError(thefile, compilerPath, silent, extraArgs):
 def dumpKeywords(compilerPath):
     '''returns tuple: (categorized-tokens-in-algebraic-form, ok-bool)'''
     import clr
-    stdout, stderr, code = launchCompiler(compilerPath, ["--listpredefined"], False)
+    silent = False
+    stdout, stderr, code = launchCompiler(compilerPath, ["--listpredefined"], silent)
```
siliconvoodoo: I hope you can see that this is a trivially useless change :)

src/AzslcMain.cpp
```diff
@@ -398,7 +398,7 @@ int main(int argc, const char* argv[])
     cli.add_flag("-r", visitRecursively, "(Option of --visitsym) Visit recursively.");
 
     bool listPredefined = false;
-    cli.add_flag("--listpredefined", "Output a list of all predefined types in AZSLang.");
+    cli.add_flag("--listpredefined", listPredefined, "Output a list of all predefined types in AZSLang.");
```
siliconvoodoo, line 401: ok. this is the real fix. good find.

siliconvoodoo (changes requested):
remove useless silent = False declaration in testfuncs.py

tests/testfuncs.py
martinwinter-huawei: good point!

tests/testfuncs.py
```diff
@@ -147,7 +147,7 @@ def dumpKeywords(compilerPath):
     import clr
     stdout, stderr, code = launchCompiler(compilerPath, ["--listpredefined"], False)
     if code != 0:
-        if not silent: print (clr.fg.RED + "compilation failed" + clr.style.RESET_ALL)
```
siliconvoodoo, line 150: I see. missed that one. good job.

siliconvoodoo (approved):
good for me

jeremyong-az (approved):

galibzon (approved):

PR #61, opened 2022-09-20
Updating to hlsl6 6

siliconvoodoo:
It's not a complete update (missing payload PAQ) but it's a big overall refresh.
List of changes:

1/
report that `globallycoherent` keyword is missing
temporarily bypassable by using `[verbatim("globallycoherent)"]` workaround.

went to
https://github.com/microsoft/DirectXShaderCompiler/blob/50c0199c17b0b7530cbc0bdddadc3697eaece695/tools/clang/include/clang/Basic/TokenKinds.def
to find missing tokens

list is incomplete, need to crossref with doc:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-keywords

missing and included:
```
center
export
globallycoherent
sampler_state
snorm
unorm
tbuffer
indices
vertices
payload
```

-> problem of reserving names. DXC uses "special identifier" semantic.
in AntlR: it's harder (refer to 2/)

example from DXC repository:  center_kwd.hlsl
```
  // RUN: %dxc -T ps_6_0 -Od -E main %s | FileCheck %s
  // CHECK: @main
  // make sure 'center' is allowed as an interpolation modifier
  float main(center float t : T) : SV_TARGET
  {
      // and also as an identifier
      float center = 10.0f;
      return center * 2;
  }
```
decided not to include yet:
```
min16float
min10float
min16int
min12int
min16uint
asm
asm_fragment
```
not to include probably ever:
```
compile_fragment
fxgroup
pixelfragment
vertexfragment
RenderTargetView, DepthStencilView, DepthStencilState
RasterizerState
VertexShader, ComputeShader, CompileShader, DomainShader, GeometryShader, HullShader, compile
technique, technique10, technique11
```
2/
Noticed while testing function declaration parsing,
mistake that dates back May 2020:
it is impossible to use "in out" or "const" or any qualifier other than static or inline on nameless function parameters, and return type.

found at random by changing parsing rules around qualifiers which caused a greedy parse and shifted param name into type.
rule is:

`qualifier* type name?`

so
"const int i"  ok because named-param
"const int"    not ok because unnamed-param zealous semantic check/bug

there may be a solution to use a generic lexer match (Identifier) inside qualifier rule.
using qualifier*? which causes the rule to be non-greedy. to investigate later.

in 1.7.35 this can't build:
```
static const int f()
{
    return 2;
}
// Semantic error #33: Functions can only have either static or inline modifiers.
```

--> fixed by inverting the logic (all passes except AZSL additions: options/rootconstant).

3/
mesh shaders

https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#example-1-passthrough

example from document doesn't build:
```
[outputtopology("triangle")]
[numthreads(NUM_THREADS_X, NUM_THREADS_Y, NUM_THREADS_Z)]
void PassthroughMesh shader(
    in uint tid : SV_DispatchThreadID,
    in uint tig : SV_GroupIndex,
    out vertices MyOutputVertex verts[MAX_NUM_VERTS],
    out indices uint3 triangles[MAX_NUM_PRIMS])
```
`PassthroughMesh` isn't a keyword it's a typo. That was hard to find because of strong subconscious assumption that the doc is the truth.

Similarly:
many official examples and tests have `SV_RayPayload` semantic.

https://github.com/microsoft/DirectXShaderCompiler/search?p=1&q=sv_raypayload

This is a fakenews semantic that doesn't exist. it's ignored. no codepath in DXC source can recognize it.

4/
Working mesh shader found in D3D12MeshShaders sample solution.

Surprise found:
`ConstantBuffer<MeshInfo>    MeshInfo            : register(b1);`

This can't pass AZSLc because MeshInfo variable has the same name as MeshInfo type.
message:
> ConstantBuffer<T>'s generic type /u_/MeshInfo must be user defined, but seen as IsNotType

same problem for
```
struct A{};
A A;
```
> redeclaration of /A with a different kind: Variable but was Struct, first seen line 1

Investigated what to do, but concluded against support. (c.f comment in code at commit 5a927ba62e7fab29df69fb74047)

5/
16 bits types.

preamble:
some person from Tencent added:
```
uint32_t
uint64_t
```
but not their vector, and matrix keyword expansions:
```
uint32_t3
uint32_t1x1
...
```

for 16bits:
I remember we added -enable-16bit-types to the builder a long time ago.
That said, there is no grammar support for int16_t
doc:
https://github.com/microsoft/DirectXShaderCompiler/wiki/16-Bit-Scalar-Types

the doc says "scalar type" but DXC actually swallows all the vector/matrix variants as well.

6/
I managed to improve the syntax error display from:
syntax error #1: no viable alternative at input 'voidmain(payloadMyPayloadpayload'
to:
syntax error #1: no viable alternative at input 'void main(payload MyPayload payload'

by using a hidden channel and a comment channel on antlr lexer.

7/
noticed but not yet fixed:
in SRGsemantic block:
`ShaderVariantFallback = stuff`

the initializer "stuff" is not well checked and can be anything (string, bool...)

8/
fixed up launch_grun.bat (problem with CLASSPATH)

9/
Since the support of `#line` directive, we reported messages with 1 line off.
This is due to a misunderstanding of `#line` behavior.

```
#line 1 "myprog"
error  // myprog:1:1: error: unknown type name 'error'
```
standard 16.4.3:
> behave as if the following sequence of source lines begins with a source line that has a line number as specified

9.5/
refactoring: the output stream sill counts linefeeds, but now does it through an interface Streamable.
the line emission system is entirely redone, and we don't emit unnecessary linefeeds anymore.

10/
there is an antlr 4.11 that could be interesting since huge changes happened in 4.10 on the C++ platform (lots of optimizations).
we have 4.9.3 now.
it seems the scheme to get it is through a git clone (called by cmake?) to an "external" o3de repository?

11/
`unsigned int16_t g;`

this crashes DXC.

> Internal compiler error: access violation. Attempted to read from address 0x0000000000000008

reported at https://github.com/microsoft/DirectXShaderCompiler/issues/4648

12/
grammar:
Because of the introduction of new int types, "unsigned" needed to move to the qualifiers rule,
to lift the necessity of repetitions of 20 vector/matrix expansions x2 in the lexer, and allow arbitrary placement eg `unsigned const int`.
This disrupted the AST possible structure, which now will tolerates qualifiers in type contexts.
effect on code:
-> refactoring: type qualifiers are no longer stored in VarInfo and FuncInfo but in ExtendedTypeInfo

example previsouly syntax errors, but now passing:
```
typedef const int UI;
ConstantBuffer< volatile Color >
typealias U = const T;
const float f();   // returning const types was not acceptable grammar in 1.7.35
```
previously semantic error but now pass unchecked:
```
ConstantBuffer< ConstantBuffer<A> > v;  // it was enforced to not allow that.
ConstantBuffer<A> func(ConstantBuffer<B> cbb)  // functions couldn't return CB nor have them as argument
{
  ConstantBuffer<A> cb = cbb;  // local variables with CB type were hard errors (only tolerated in SRG)
}
```
rationale: there is an opportunity for DXC to recognize references types as aliases (that have no physical existence)
For example when used in selector Functions
`ConstantBuffer<S> Select(bool a_or_b) { return a_or_b ? g_cb1 : g_cb2; }`
Even though DXC sort of seem to build this, in some contexts it does trigger a complaint that this type
shouldn't appear here.
To be fair, the selector function can undecorate the viewtype (chameleon type in azslc parlance) by performing the access:
`S Select(bool a_or_b) { return a_or_b ? g_cb1 : g_cb2; }`
and be more guaranteed to be legal.
But anyhow, I decided we don't need to care to validate that, let's reduce code complexity by offsetting burden of validation to downstream tools.

13/
Further improvement of syntax error reporting by detecting when the offending token is a keyword.
this way, if someone gets a "no viable alternative" on a visibly valid code:
`void f(Payload payload){}   // <- this is an error in AZSL 1.8 because payload is reserved!`
message now:
> syntax error #1: no viable alternative at input 'void f(Payload payload' (payload is a keyword)

this will relax the frustration from surprising syntax errors with supercommon names:
"indices, vertices, center, line, payload..."

14/
Syntax still remaining to support:
PAQ (HLSL 6.6 payload qualifiers)
eg:
`float4 irradiance : read(caller) : write(closesthit, miss);`

15/
Improved "option for int must use range" error message with an example syntax to teach users the solution without google.


src/AzslcMain.cpp
```diff
@@ -130,6 +139,17 @@ namespace AZ::ShaderCompiler
...
     }
 
+    bool IsKeyword(const antlr4::Recognizer* r, antlr4::Token* token)
+    {
+        MapOfStringViewToSetOfString byTypeClass;
+        set<string> notTypes;
+        VisitTokens(r, byTypeClass, notTypes);
+        bool notType = notTypes.find(token->getText()) != notTypes.end();
+        bool notIdentifier = r->getVocabulary().getSymbolicName(token->getType()) != "Identifier";
+        bool firstIsalpha(token->getText()[0]);
```
siliconvoodoo: ah. I forgot `isalpha` function here. I'll include it in the next PR.

src/AzslcMain.cpp
```diff
@@ -480,14 +502,21 @@ int main(int argc, const char* argv[])
...
         parser.removeErrorListeners();
+        azslParserEventListener.m_isKeywordPredicate = IsKeyword;
         parser.addErrorListener(&azslParserEventListener);
         tree::ParseTree *tree = parser.compilationUnit();
 
+        if (ast)
+        {
+            PrintAst(tree, parser);
+            syntax = true; // ast print is a syntax only build.
+        }
```
siliconvoodoo, line 518: moving this up here allows to dump ast even in case of semantic errors.

src/AzslcMain.cpp
```diff
@@ -466,7 +485,10 @@ int main(int argc, const char* argv[])
         }
 
         const string inputFileName = useStdin ? "" : inputFile;
-        AzslcException::s_currentSourceFileName = useStdin ? "stdin" : inputFile;
+        PreprocessorLineDirectiveFinder lineFinder;
```
siliconvoodoo, line 488: the line finder has been extracted from the IR. it nows lives its own independent life, and is pointer-passed to a static field of the base exception class. This way it cleaned up the static field storing the original file name.

src/AzslcPredefinedTypes.h
```diff
@@ -165,25 +165,27 @@ static constexpr std::array<const char*, 2> OtherPredefined = {
...
 "half",
 "int",
+"int16_t",
 "int32_t",
 "int64_t",
 "uint",
+"uint16_t",
 "uint32_t",
-"uint64_t",
-"unsigned int"};
```
siliconvoodoo, line 186: `unsigned int` is not a hard `unsigned⎵int` anymore but a combo of qualifier `unsigned` and core type `int`

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -486,28 +497,34 @@ namespace AZ::ShaderCompiler
...
+        auto uqNameView                 = UnqualifiedNameView{idText};
+        // Register the variable in the symbol table early:
+        IdAndKind& symbolRef            = AddIdentifier(uqNameView, Kind::Variable, line);
+        auto& [uid, kindInfo]           = symbolRef;
+        VarInfo& varInfo                = kindInfo.GetSubRefAs<VarInfo>();
+        // Discover array dimensions:
         ArrayDimensions arrayDims;
         TryFoldArrayDimensions(ctx, arrayDims);
-        // discover matrix majorness
-        Packing::MatrixMajor major = ExtractMatrixMajorness(varInfo);
```
siliconvoodoo, line 507: no need to do that as a special case anymore. this is done canonically by CreateExtendedTypeInfo.

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -1172,11 +1186,6 @@ namespace AZ::ShaderCompiler
         return LookupType(ctx).GetName();
     }
 
-    QualifiedName SemanticOrchestrator::TypeofExpr(AstFuncType* ctx) const
```
siliconvoodoo, line 1175: AstFuncType doesn't exist anymore. it got fused inside of AstType. So there is no more two grammar rules, one that can accept "void" and one that can't. The only rule left is all powerful. The only point this had, and it came all the way back from the Tim Jones .g4 version, is that it was forbidding natively, by grammar check, the possibility of declaring void variables. We don't need that strictness, especially not at the obscure message-reporting grammar level.

src/AzslcUtils.h
```diff
@@ -1087,86 +1110,44 @@ namespace AZ::ShaderCompiler
...
-             : ctx->Nointerpolation() ? StorageFlag::Unknown
-             : ctx->Noperspective()   ? StorageFlag::Unknown
-             : ctx->Sample()          ? StorageFlag::Unknown
-             : ctx->Point()           ? StorageFlag::Unknown
-             : ctx->Line_()           ? StorageFlag::Unknown
-             : ctx->Triangle()        ? StorageFlag::Unknown
-             : ctx->LineAdj()         ? StorageFlag::Unknown
-             : ctx->TriangleAdj()     ? StorageFlag::Unknown
-             :                          StorageFlag::Unknown;
+        return AsFlag(ctx) == flag;
```
siliconvoodoo, line 1149: simplified to one line. less maintenance. more better.

src/GenericUtils.h
```diff
@@ -18,6 +19,37 @@
...
+        if (object)
+        {
+            return functor(object);
+        }
+        else
+        {
+            return VisitFirstNonNull(functor, rest...);
+        }
+    }
+
```
siliconvoodoo, line 52: migrated that over from StdUtils. It had nothing to do in that header. StdUtils is for std configuration on platforms, usings, and paliate shortcomings of versions/API of std things only.

src/GenericUtils.h
```diff
@@ -212,14 +249,15 @@ namespace AZ
     template< typename Iter >
     string Join(Iter begin, Iter end, string_view separator = "")
     {
-        if (begin == end)
+        if (!(begin != end))
```
siliconvoodoo, line 252: this is to tolerate Joining of ReflectableEnums::Enumerate{} which has a super tiny iterator that only has `!=`

src/GenericUtils.h
```diff
@@ -212,14 +249,15 @@ namespace AZ
...
             return "";
 
         std::stringstream ss;
-        ss << *begin;
+        Streamable&& wrap{MakeOStreamStreamable{ss}};
+        wrap << *begin;
 
-        auto aggregate = [&ss, &separator](auto s) { ss << separator.data() << s; };
-        std::for_each(std::next(begin), end, aggregate);
```
siliconvoodoo, line 222: for some reason I couldn't make next() work on ReflectableEnum::Iterator. I'm guessing prefix++ will be more compatible anyway?

src/StreamableInterface.h
```diff
@@ -0,0 +1,48 @@
...
+    public:
+        virtual Streamable& operator<<(char) = 0;    // template virtual method are forbidden in C++, so manual listing :(
+        virtual Streamable& operator<<(const char*) = 0;
+        virtual Streamable& operator<<(double) = 0;
+        virtual Streamable& operator<<(int64_t) = 0;
+        virtual Streamable& operator<<(uint32_t) = 0;
+        virtual Streamable& operator<<(size_t) = 0;
+        virtual Streamable& operator<<(bool) = 0;
+        virtual Streamable& operator<<(const std::string&) = 0;
+    };
```
siliconvoodoo, line 28: I didn't want template contamination in the Emitter class hierarchy, so virtual calls would be my alternative. It ended up needing lots of this boilerplate that I highly dislike. If we can find a better solution in the future that'd be neat. Maybe accept the template creep...

src/Texture2DMSto2DCodeMutator.cpp
```diff
@@ -336,18 +336,35 @@ namespace AZ::ShaderCompiler
...
-        auto typeCtx = ExtractTypeFromVariableDeclarator(ctx, &paramCtx);
-        auto variableTypeStr = typeCtx->getText();
+        auto typeCtx = ExtractTypeFromUnnamedVariableDeclarator(ctx, &paramCtx);
+        auto tokens = NodeTokens(typeCtx, stream);
+        vector<string> stringlets;
+        TransformCopy(tokens, stringlets, [&](Token* t) { return t->getText(); });
+        vector<string> filtered;
+        std::copy_if(stringlets.begin(), stringlets.end(), std::back_inserter(filtered), [&](auto subtok)
+                     { return subtok != "in" && subtok != "out" && subtok != "inout" && !IsAllWhitespaces(subtok); });
+        string typeHlsl = Join(filtered, " ");
```
siliconvoodoo, line 362: oh the joy this will be in C++20.

tests/Emission/output-attribute.txt
```diff
@@ -0,0 +1 @@
+" # pragma OutputFormatHint ( target 0 R32 ) "
```
siliconvoodoo, line 1: because I had a bug where the Streamable would shove an integer typer into an `char` and 0 was interpreted as \0 which was interesting. DXC can swallow null characters, it shows a simple warning, funny.

tests/Semantic/AsError/disallowed-function-modifiers.azsl
```diff
@@ -1,21 +1,3 @@
...
-centroid void FuncCD();
-nointerpolation void FuncNI();
-noperspective void FuncNP();
-sample void FuncSpl();
-in void FuncIn();
-out void FuncOut();
-inout void FuncIO();
+option void Func1(); // #EC 33
 
-// Semantic Error 33: line 1 ' Functions can only have the static and inline modifiers!'
```
siliconvoodoo, line 20: that was a bad idea. we need to allow at least const, unorm, snorm, precise, matrix stuff etc. So instead of being strict, let's be loose.

src/AzslcBackend.cpp
```diff
@@ -703,17 +705,60 @@ namespace AZ::ShaderCompiler
...
+        bool isMatrix = typeInfo.m_coreType.m_arithmeticInfo.IsMatrix();
+        if (typeInfo.CheckHasStorageFlag(StorageFlag::ColumnMajor) && !(bannedFlags & StorageFlag::ColumnMajor))
+        {
+            modifiers = "column_major";
+        }
+        else if (typeInfo.CheckHasStorageFlag(StorageFlag::RowMajor) && !(bannedFlags & StorageFlag::RowMajor))
+        {
+            modifiers = "row_major";
+        }
+        else if (options.m_forceEmitMajor && isMatrix)
```
jeremyong-az, line 722: nit: I think I might not know what `m_forceEmitMajor` would have referred to if looking at the options alone (away from the context here). Maybe `m_forceMatrixLayout` and `m_forceMatrixRowMajor`?

src/AzslcEmitter.cpp
```diff
@@ -460,23 +463,35 @@ namespace AZ::ShaderCompiler
...
-        m_alreadyEmittedPreprocessorLineDirectives.insert(physicalTokenLine);
+        if (azslLineNumber == 0)
+            return;  // protect for this invalid case. seems to happen for "virtual" symbols (like OverloadSet)
+
+        size_t supposedVirtualLine = m_lineFinder->GetVirtualLineNumber(azslLineNumber);
+        size_t curHlslLine = m_out.GetLineCount() + 1;  // "lines" is a space that is 1-based indexed.
+        auto lastEmitted = Infimum(m_alreadyEmittedPreprocessorLineDirectives, curHlslLine);
+        if (lastEmitted != m_alreadyEmittedPreprocessorLineDirectives.cend())
+        {
+            // verify if we can skip the line emission if current stream line feed is still in synch with expectations
```
jeremyong-az: Nit: "sync"

src/AzslcEmitter.cpp
```diff
@@ -460,23 +463,35 @@ namespace AZ::ShaderCompiler
...
-        {
-            return;
-        }
-
-        const string& originalFileName = StdFs::absolute(lineDirectiveInfo->m_containingFilename).lexically_normal().generic_string();
-        const auto lineNumber = lineDirectiveInfo->m_forcedLineNumber;
-        m_out << "#line " << lineNumber << " \"" << originalFileName << "\"\n";
-
-        m_alreadyEmittedPreprocessorLineDirectives.insert(physicalTokenLine);
+        if (azslLineNumber == 0)
```
jeremyong-az, line 466: This is a great fix thanks. Thinking out loud, we may want an option to strip the `#line` directives altogether in situations where we aren't debugging a shader so that changes to a stripped function (once we have function stripping) won't affect line numbers that affect the file hash. Alternatively, in the future perhaps we implement a custom CRC that bypasses the line directives. At the cost of a slightly slower CRC

src/AzslcEmitter.h
```diff
@@ -211,7 +212,7 @@ namespace AZ::ShaderCompiler
         SymbolTranslation m_translations;
         unordered_set<IdentifierUID> m_alreadyEmittedFunctionDeclarations;
         unordered_set<IdentifierUID> m_alreadyEmittedFunctionDefinitions;
-        unordered_set<size_t> m_alreadyEmittedPreprocessorLineDirectives;
+        map<size_t, size_t> m_alreadyEmittedPreprocessorLineDirectives;
```
jeremyong-az, line 215: Suggest `unordered_map` since it doesn't look like ordered traversal is needed

src/AzslcMain.cpp
```diff
@@ -32,15 +34,14 @@ namespace AZ::ShaderCompiler
 
     using MapOfStringViewToSetOfString = map<string_view, set<string>>;
 
-    // out argument: classifiedTokens
-    // filter: is a predicate for condition to check to pass registration
-    template <typename FilterFunction>
-    void ClassifyAllTokens(const azslLexer* lexer, MapOfStringViewToSetOfString& classifiedTokens, FilterFunction&& filter = nullptr)
+    template <typename TypeClassFilterPredicate = std::nullptr_t>
+    void VisitTokens(const antlr4::Recognizer* recognizer,
+                     MapOfStringViewToSetOfString& acceptedToken, set<string>& notTypes1,  // out
```
jeremyong-az, line 39: `notTypes1` appears unused in this function?

src/AzslcSemanticOrchestrator.cpp
```diff
@@ -128,19 +141,19 @@ namespace AZ::ShaderCompiler
...
         if (!originalDeclarationAsFunc)
         {
             Kind realKindOfOriginallyDeclaredMember = m_symbols->GetIdAndKindInfo(optionalIdentifier->GetName())->second.GetKind();
-            ThrowAzslcOrchestratorException(ORCHESTRATOR_UNEXPECTED_KIND,
+            throw AzslcOrchestratorException{ORCHESTRATOR_UNEXPECTED_KIND,
                 ctx->start, ConcatString((scopeIsClass ? "class " : "SRG "), scopeUid.m_name,
                      " holds a member ", optionalIdentifier->m_name,
                      " but it is of kind ", string{ Kind::ToStr(realKindOfOriginallyDeclaredMember) },
-                     " instead of expected ", string{ Kind::ToStr(Kind::Function) }));
+                                         " instead of expected ", string{ Kind::ToStr(Kind::Function) })};
```
jeremyong-az, line 156: Nit: whitespace in this file feels a bit off. Does this project have a clangformat?

galibzon (approved):

jeremyong-az (commented):
(will continue to review just leaving a few comments for now, thanks!)

src/AzslcUtils.h
```diff
@@ -273,9 +258,36 @@ namespace AZ::ShaderCompiler
...
+            vector<StorageFlag> bag;
+            auto end = std::copy_if(StorageFlag::Enumerate{}.begin(), StorageFlag::Enumerate{}.end(), std::back_inserter(bag),
+                                    [&](auto sf) -> bool { return (m_flag & sf) && (sf & ~StorageFlag::Other); });
+            return string{Trim(Join(bag.begin(), bag.end(), " ") + " " + Join(m_others.begin(), m_others.end(), " "))};
+        }
+
+        void OrMerge(const TypeQualifiers& src)
+        {
+            m_flag |= src.m_flag;
+            set<string> fresh{m_others.begin(), m_others.end()};
```
jeremyong-az: Suggest `unordered_set`

src/PreprocessorLineDirectiveFinder.h
```diff
@@ -29,18 +29,53 @@ namespace AZ::ShaderCompiler
     class PreprocessorLineDirectiveFinder
     {
     public:
-        virtual ~PreprocessorLineDirectiveFinder() {};
+        const LineDirectiveInfo* GetNearestPreprocessorLineDirective(size_t physicalLine) const
+        {
+            auto it = Infimum(m_lineMap, physicalLine);
```
jeremyong-az, line 34: I didn't expect to see terms from topology in this code review! :)

src/StreamableInterface.h
```diff
@@ -0,0 +1,48 @@
...
+* SPDX-License-Identifier: Apache-2.0 OR MIT
+*
+*/
+#pragma once
+
+#include <ostream>
+
+namespace AZ
+{
+    class Streamable
```
jeremyong-az, line 17: A comment on how this type/concept is used or expected to behave might be useful

tests/Emission/ray-tracing-procedural-geometry.azsl
```diff
@@ -323,37 +319,37 @@ bool SolveQuadraticEqn(float a, float b, float c, out float x0, out float x1)
...
+    return normalize(hitPosition - a_center);
 }
 
 
 
-bool SolveRaySphereIntersectionEquation(in Ray ray, out float tmin, out float tmax, in float3 center, in float radius)
+bool SolveRaySphereIntersectionEquation(in Ray ray, out float tmin, out float tmax, in float3 a_center, in float radius)
 {
-    float3 L = ray.origin - center;
+    float3 L = ray.origin - a_center;
```
jeremyong-az, line 332: `center` as a keyword, while present in the DXC tokenizer is something I'm not sure we really should support as it's far too common in everyday shader code IMO. If ANTLR struggles to tokenize it, I suggest we just not support the keyword for now (I'm not actually sure what it means in HLSL as a keyword)

src/AzslcEmitter.h
siliconvoodoo: this one needs order for upper_bound lookup

src/AzslcSemanticOrchestrator.cpp
siliconvoodoo: right, that's a merge issue that happened because of the "signed-off" rebase of the other branch

src/AzslcUtils.h
siliconvoodoo: this one indeed

src/PreprocessorLineDirectiveFinder.h
siliconvoodoo: 😛

tests/Emission/ray-tracing-procedural-geometry.azsl
siliconvoodoo: We'll do an improvement with the `qualifier*?` rule later using a lazy Identifier, This should allow us to delete a bunch of keywords. center is used in the same context as centroid according to the online docs. I suggest waiting for the next PR to be able to move on first, then get the improvements.

src/AzslcEmitter.cpp
siliconvoodoo: We should prefilter the file contents in the builders instead no?
We can run a regexp that remove dead decoration like comments and line directives, and/or maybe a formatter (astyle/clang/treesitter?) with the explicit goal of producing very stable and deterministic outputs.

src/AzslcMain.cpp
siliconvoodoo: it's used line 60

tests/Emission/ray-tracing-procedural-geometry.azsl
siliconvoodoo: victory. this idea worked!

jeremyong-az (approved):

siliconvoodoo:
all green. good to go.

PR #65, opened 2022-10-19
Fix pad_to doesn't silence error IR 131

siliconvoodoo:
In case of such a program as input:

```
ShaderResourceGroupSemantic ExampleBinding { FrequencyId = 0; };
ShaderResourceGroup ExampleSRG : ExampleBinding
{
    float3x3 mat;
    bool b;
};
```

We get the result

> IR error #131: Detected potential alignment issues related with DXC flag '-fvk-use-dx-layout'.
> Alternatively you can use the option '--no-alignment-validation' to void this error
> - A 'float2' variable should be added before the variable 'b'

It seems natural that we can either to the recommendation, or use the `[[pad_to()]]` attribute as such:

```
    float3x3 mat;
    [[pad_to(44)]]
    bool b;
```

Doing so in conjunction with --no-alignment-validation did result in a float2 being inserted so the error message should go away. But it didn't. This PR fixes that.

Secondly, there was a crash in case of this input:
```
    float2x2 mat;
    [[pad_to(36)]]
    bool b;
```


siliconvoodoo:
problem with incompatible history caused by squashed branch. reopened as #66

PR #68, opened 2022-10-20
Fix crash

siliconvoodoo:
...uppon simple redeclaration as different kind when it's a function.


siliconvoodoo:
@santorac @galibzon ping? one line change :)

galibzon (approved):

rgba16f (approved):

PR #74, opened 2022-12-07
little fix: `static` in SRG caused double emission

siliconvoodoo:
prohibit static SRG members emission, because static members don't participate in the resources, they are normal local variables. they will be taken care of by their normal emission path.

+ add possiblity to test for the ABSCENCE of code in emssion tests using ^ character at beginning of line (as for [^] in regex lang).

input:
```c
ShaderResourceGroup S : slot1
{
    static Texture2D t;
};
```
result before:
```c
static Texture2D S_t : register(t0, space0);
static Texture2D S_t;
```
result now:
```c
static Texture2D S_t;
```


galibzon (approved):

siliconvoodoo:
@VickyAtAZ ping for review? The idea here is that shader programmers who use "static" in SRG scope (for example to store pseudo global states) had a bug that caused invalid HLSL. Now fixed.

akioCL (approved):

PR #76, opened 2023-01-12
Fixing order of multiple unbounded arrays when unique indices are used.

invertednormal:
Previously it was assumed that only one unbounded array would be used, but this is no longer true. That caused the order of the unbounded arrays to be shuffled (the first unbounded array would be moved to the end) and other unbounded arrays might not appear at the end of the binding table. This would then cause validation failures in Vulkan.

This fixes it so that all unbounded arrays appear together at the end, in the original order they were originally defined.

Ran all AZSLC tests, tested in AtomSampleviewer's bindless sample, and tested with terrain using bindless textures.

### DIFF
--- a/src/AzslcBackend.cpp
+++ b/src/AzslcBackend.cpp
@@ -636,7 +636,7 @@ namespace AZ::ShaderCompiler
 
             for (const auto tId : srgInfo->m_srViews)
             {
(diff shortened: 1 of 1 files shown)


moudgils (approved):

rgba16f (approved):

cgalvan (approved):

siliconvoodoo (approved):

invertednormal:
Fixes [#13813](https://github.com/o3de/o3de/issues/13813)

PR #77, opened 2023-01-15
Fix for "option range not reconstructed"

siliconvoodoo:
report github.com/o3de/o3de-azslc/issues/75

This adds the min by using a id (name) reference to the first enumerator name, in case of enums, or it emits a full switch in case the enum is complicated. In case of integer range, it reproduces the value by adding the min as an integrer literal value from the range attribute. I noted that there is no intelligence in reproduction of the underlying type, it's always uint, this will probably be an issue for negative range int options. Need to open a second issue after a bit of investigation about this.

The test suite has a new test to check this new emission pattern.
the test suite tests green on windows.

tests not done: in runtime (output colors according to option retrieved from constant buffer key (must use root variant))


galibzon (approved):

siliconvoodoo:
DXC won't build. Must do some casts. On it now.
image

siliconvoodoo:
I found a local hack to empirically test the viability of the change.

By changing AuxGeomObject.azsl's pixel shader to:

```
[[range(4,8)]]
option int o_clr = 5;

PSOutput MainPS()
{
    PSOutput OUT;
    OUT.m_color = ObjectSrg::m_color;
    if (o_clr == 5)
        OUT.m_color = float4(1,0,1,1);
    else if (o_clr == 6)
        OUT.m_color = float4(1,1,0,1);
    else if (o_clr == 7)
        OUT.m_color = float4(1,0,0,1);
    return OUT;
}
```
and adding
```
shaderOptionAndValues.push_back(
                    RPI::ShaderOption(Name("o_clr"), Name("6")));
```
in FixedShapeProcessor.cpp line 1515.
and
```
console->PerformCommand(AZStd::string::format("r_forceRootShaderVariantUsage %s", "true").c_str());
```
In ProceduralSkinnedMeshComponent.cpp

We get
image

I verified that setting 7 gets red, setting nothing gets majenta.
And that using azslc 1.8.12 (prior to fix) even when the option is set to "6" we get blue, which is the object color.
image

Which proves the problem existed, and the fix was necessary.

src/AzslcEmitter.cpp
```diff
@@ -1140,14 +1140,68 @@ namespace AZ::ShaderCompiler
...
+                if (otherInitializers)
+                {
+                    m_out << ";\n    switch (shaderKey)\n    {\n";
+                    for (int i = 0; i < enumerators.size(); ++i)
+                    {
+                        m_out << "        case " << i << ": shaderKey = (" << GetTranslatedName(returnType, UsageContext::ReferenceSite) << ")"
+                              << GetTranslatedName(enumerators[i], UsageContext::ReferenceSite) << "; break;\n";
+                    }
+                    m_out << "    }";
+                }
```
invertednormal, line 1186: It seems like this would force a switch statement in cases where someone was being pedantic and assigning incrementing values to the enum, ie
```
option enum class Foo {
    Foo1 = 0,
    Foo2 = 1,
    Foo3 = 2,
}
```
Would it be that much more work to just check for non-incrementing enumeration initializers?

invertednormal (approved):
Looks good, just one minor suggestion if it's not to much work.

src/AzslcEmitter.cpp
siliconvoodoo: interesting! I agree that we can make this check. Let's just do it in a separated PR if you don't mind? I'll flush it ASAP. (we can also use a `static uint` array instead of a switch i'm not sure which one the shader compilers will like better for optimization).
In the meantime we free the schedule for package with the current state of things if that's ok with you

invertednormal: yeah, definitely not a blocker, just an idea to make it better :) feel free to submit as-is.

PR #80, opened 2023-02-24
fix a runtime error in function semantic registration

siliconvoodoo:
While investigating if https://github.com/o3de/o3de-azslc/issues/37 was really fixed, I stumbled uppon remaining issues.
The fix of december https://github.com/o3de/o3de-azslc/commit/0c4eccd90108f6283a813f996e24acb4f6635364
definitely fixed a crash, but it wasn't investigated further after that. In fact in unlocked the door to at least 2 scenarios that don't finish compilation in a normal manner (fatal internal).
I added 2 tests to make sure we don't regress on those 2. And the simplest fix I could find, in the form of a mini 3 state automata in the function symbol info.

before:
case 1:
image
(no code output on stdout since fatal error interrupted)
case 2:
image
(no code output, only stderr ODR message)

after:
image
Same for both 2 cases.

### DIFF
--- a/src/AzslcListener.cpp
+++ b/src/AzslcListener.cpp
@@ -221,13 +221,17 @@ namespace AZ::ShaderCompiler
 
     void SemaCheckListener::exitFunctionParam(azslParser::FunctionParamContext* ctx)
     {   // we use the exit rule to let the time to inlined-UDT-decl to get registered through the type rule visit first.
-        if (!ctx->Name)
+        auto& funcInfo = m_ir->m_sema.GetCurrentScopeSubInfoAs<FunctionInfo>();
+        if (funcInfo.m_multiFwds != FMF_FwdDeclRedundancy)  // when in that state, we don't accept parameter registration
         {
-            m_ir->m_sema.RegisterNamelessFunctionParameter(ctx);
-        }
-        else
-        {
-            m_ir->m_sema.RegisterVar(ctx->Name, ctx->unnamedVariableDeclarator());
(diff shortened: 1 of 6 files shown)


galibzon (approved):

src/AzslcListener.cpp
```diff
@@ -221,13 +221,17 @@ namespace AZ::ShaderCompiler
 
     void SemaCheckListener::exitFunctionParam(azslParser::FunctionParamContext* ctx)
     {   // we use the exit rule to let the time to inlined-UDT-decl to get registered through the type rule visit first.
-        if (!ctx->Name)
+        auto& funcInfo = m_ir->m_sema.GetCurrentScopeSubInfoAs<FunctionInfo>();
+        if (funcInfo.m_multiFwds != FMF_FwdDeclRedundancy)  // when in that state, we don't accept parameter registration
```
rgba16f, line 225: Does AZSL support function overloads? And would this break it?

rgba16f (approved):

src/AzslcListener.cpp
siliconvoodoo: Excellent question. To which I got a bit startled, there is a real possibility that it could.
In theory, there is an isolation between overloads from their signature which includes parameters, so the "GetCurrentScope..." function should be able to keep states pinpointed to each overload.

So, I just went and did a small check, to verify if there is a problem.
image
This is good, seems that the HLSL is as expected.

And symbol table looks as expected too
image

phew. seems that we're good.

PR #85, opened 2023-04-27
Auto option ranks (improvement)

siliconvoodoo:
Intermediary PR update, in case the investigation further into light costs finds nothing special, this is a first shot at the better evaluation results.

### DIFF
--- a/src/AzslcReflection.cpp
+++ b/src/AzslcReflection.cpp
@@ -1052,52 +1052,80 @@ namespace AZ::ShaderCompiler
             // only options
             if (varInfo->CheckHasStorageFlag(StorageFlag::Option))
             {
+                verboseCout << "Analyzing " << uid << "\n";
                 int impactScore = 0;
                 // loop over appearances over the program
                 for (Seenat& ref : kindInfo->GetSeenats())
                 {
+                    verboseCout << "Seenat line " << ref.m_where.m_line << "\n";
                     // determine an impact score
                     impactScore += AnalyzeImpact(ref.m_where)  // dependent code that may be skipped depending on the value of that ref
                         + 1;  // by virtue of being mentioned (seenat), we count the reference as an access of cost 1.
                 }
                 varInfo->m_estimatedCostImpact = impactScore;
+                verboseCout << uid << " final cost " << impactScore << "\n";
             }
         }
+    }
+
+    template< typename CtxT >
+    bool SetNextNodeIfPartOfTypeCtxTCondViaNParents(
+        ParserRuleContext*& node,
+        int maxDepth)
+    {
+        if (auto* searchNode = DeepParentAs<CtxT*>(node, maxDepth))
+        {
+            if (IsParentOf(searchNode->Condition, node))
+            {
+                node = searchNode->embeddedStatement();
+                return true;
+            }
+        }
+        return false;
     }
 
     int CodeReflection::AnalyzeImpact(TokensLocation const& location) const
     {
         // find the node at `location`:
         ParserRuleContext* node = m_ir->m_tokenMap.GetNode(location.m_focusedTokenId);
-        // go up tree to meet a block node that has visitable depth:
-        // can be any of if/for/while/switch
-        //  4 is an arbitrary depth, enough to search up things like `for (a, b<(ref+1), c)` binaryop->braces->cmpexpr->cond->for
-        if (auto* whileNode = DeepParentAs<azslParser::WhileStatementContext*>(node->parent, 3))
+        // the "belonging" statements that we will consider, before recursing:
+        using AstIf     = azslParser::IfStatementContext;
+        using AstFor    = azslParser::ForStatementContext;
+        using AstWhile  = azslParser::WhileStatementContext;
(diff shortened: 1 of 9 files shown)


src/AzslcReflection.cpp
```diff
@@ -1052,52 +1052,80 @@ namespace AZ::ShaderCompiler
...
     }
 
+    template< typename CtxT >
+    bool SetNextNodeIfChildOfCtxTCondViaNParents(
+        ParserRuleContext*& node,
+        int maxDepth)
+    {
+        if (auto* searchNode = DeepParentAs<CtxT*>(node, maxDepth))
+        {
+            if (IsParentOf(searchNode->Condition, node))
```
siliconvoodoo, line 1078: this is a fix of a "flaw" that the previous code had. this new check prevents from linking references to unrelated blocks, as such:
```
if (unrelated_condition)
{
   o_reference_close_to_if_but_not_inside;
}
```

src/AzslcReflection.cpp
```diff
@@ -1052,52 +1052,80 @@ namespace AZ::ShaderCompiler
...
+        // the "belonging" statements that we will consider, before recursing:
+        using AstIf     = azslParser::IfStatementContext;
+        using AstFor    = azslParser::ForStatementContext;
+        using AstWhile  = azslParser::WhileStatementContext;
+        using AstDo     = azslParser::DoStatementContext;
+        using AstSwitch = azslParser::SwitchStatementContext;
+        // go up the tree to meet one of them using arbitrary max depths of {5,6,7},
+        // just enough to search up things like `for (a, b<(ref+1), c)` idExpr->IdentifierExpression->OtherExpression->BinaryExpr->ParenthesisExpr->BinaryExpr->Condition->For
+        int complexityFactor = 1;
+        bool isNonLoop = SetNextNodeIfChildOfCtxTCondViaNParents<AstIf>(node, 6);
```
siliconvoodoo, line 1100: Now that we have the "unrelated AST node check" we can increase the search depth without being afraid of catching false positives. So, now bumped to "6" for `if` node for instance.

moudgils (approved):

siliconvoodoo:
A story that comes with the feature for deeper understanding https://github.com/o3de/o3de-azslc/wiki/Options-cost-estimates

src/AzslcUtils.h
```diff
@@ -958,6 +958,19 @@ namespace AZ::ShaderCompiler
...
     }
 
+    //! Verify filiation of an AST rule
+    inline bool IsParentOf(tree::ParseTree* assumedParent, tree::ParseTree* assumedChild)
+    {
+        tree::ParseTree* parent =  assumedChild->parent;
+        while (parent)
+        {
+            if (parent == assumedParent)
+                return true;
```
VickyAtAZ, line 969: nit:
```
{
    return true;
}
```

src/AzslcReflection.cpp
```diff
@@ -1169,12 +1192,14 @@ namespace AZ::ShaderCompiler
...
                     if (funcInfo->m_costScore == -1)  // cost not yet discovered for this function
                     {
+                        verboseCout << " " << concrete << " non-memoized. discovering cost\n";
                         funcInfo->m_costScore = 0;
                         using AstFDef = azslParser::HlslFunctionDefinitionContext;
                         AnalyzeImpact(polymorphic_downcast<AstFDef*>(funcInfo->m_defNode->parent)->block(),
                                       funcInfo->m_costScore);  // recurse and cache
                     }
                     scoreAccumulator += funcInfo->m_costScore;
+                    verboseCout << " " << concrete << " call score " << funcInfo->m_costScore << " added\n";
```
VickyAtAZ: Should "call score" be "cost score"?

src/AzslcReflection.cpp
```diff
@@ -1052,52 +1052,80 @@ namespace AZ::ShaderCompiler
             // only options
             if (varInfo->CheckHasStorageFlag(StorageFlag::Option))
             {
+                verboseCout << "Analyzing " << uid << "\n";
                 int impactScore = 0;
                 // loop over appearances over the program
                 for (Seenat& ref : kindInfo->GetSeenats())
                 {
+                    verboseCout << "Seenat line " << ref.m_where.m_line << "\n";
```
VickyAtAZ: "Seen at line"

VickyAtAZ (approved):

src/AzslcUtils.h
siliconvoodoo: roger that

src/AzslcReflection.cpp
siliconvoodoo: both are true but cost is more true. changing to that.

src/AzslcReflection.cpp
siliconvoodoo: Actually this one is a "vocabulary name" :p . It's a sort of thing that belongs in 'azslc culture' if you want. Looks like a typo, but actually not, hehe :) Since it's confusing, i'll use a middle ground: a dash.

PR #86, opened 2023-10-10
Fix a crash

siliconvoodoo:
in impact analysis caused by functions with no definition bodies.

This was discovered while working on shader management console, because of working with the latest version of standard PBR that has a late-user-implementation GetDepth function. No bodies are tolerated, so this PR fixes the fact that activating cost impact analysis makes azslc crash on them. This is not critical since the azslc binary that does that isn't live yet if I recall correctly.


VickyAtAZ (approved):

moudgils (approved):

siliconvoodoo:
just version bump added after review. merging.

PR #88, opened 2024-06-07
Add support for Specialization constants for shader options

akioCL:
Add new command line argument ("--sc-options") was added to support shader options as specialization constants. Implementation was added for Vulkan, Metal and DX12. For Vulkan (and Metal) it translates shader options directly as specialization constants using a DXC attribute [[vk::constant_id]]. For DX12 it uses a volatile constant that will later be patched using another tool.

The [[no_specialization]] attribute was added to disable using specialization constants for a specific shader option.

Based on RFC https://github.com/o3de/sig-graphics-audio/issues/163

Other relevant PRs:
https://github.com/o3de/DirectXShaderCompiler/pull/4
https://github.com/o3de/o3de/pull/18019


galibzon (approved):

moudgils (approved):
Do you not need to edit the grammar to allow specialization constant tags?

akioCL:
> Do you not need to edit the grammar to allow specialization constant tags?

No need for new attributes.

